Issue #225, opened 2026-04-14
[FEATURE] Enable Auto-Scrolling (panning) on Grids


ClaudioESSilva:
### Which component(s) does this affect?

- [x] Desktop App
- [ ] CLI Tool
- [ ] SSMS Extension
- [ ] Plan Analysis Rules
- [ ] Documentation

### Problem Statement

To navigate in the grids, we can use the scroll for an up/down movement; however, we need to use the horizontal scroll bar to move left/right.

### Proposed Solution

Implement Auto-Scrolling.

By doing "middle-click" (aka middle mouse button or scroll-wheel click), if you keep pressing it, it changes the cursor, allowing you to pan/panning (scroll by moving the mouse up or down and left or right.

https://github.com/user-attachments/assets/d80bd8da-8c99-46ec-91bc-d2fa1005d259

### Use Case

Faster navigation on the grid. Especially to scroll to the right/left.

### Alternatives Considered

N/A

erikdarlingdata:
Closed via #227

PR #226, opened 2026-04-14
implements #225

ClaudioESSilva:
## What does this PR do?

Implements #225 

- [ ] Tests
- [ ] Documentation

## How was this tested?

1. Connect to Query Store
2. Use scroll-wheel click to navigate the results grid

Check the video on #225 

## Checklist

### DIFF
--- a/src/PlanViewer.App/Controls/PlanViewerControl.axaml.cs
+++ b/src/PlanViewer.App/Controls/PlanViewerControl.axaml.cs
@@ -141,6 +141,7 @@ public partial class PlanViewerControl : UserControl
         var layoutTransform = this.FindControl<Avalonia.Controls.LayoutTransformControl>("PlanLayoutTransform")!;
         _zoomTransform = (ScaleTransform)layoutTransform.LayoutTransform!;
 
+        Helpers.DataGridBehaviors.Attach(StatementsGrid);
     }
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


erikdarlingdata (commented):
Nice work — clean separation with the `DataGridBehaviors` helper, and handling Avalonia's scrollbar quirks correctly. A few things I noticed:

**1. Reflection on internal methods (fragile)**

`ProcessHorizontalScroll` / `ProcessVerticalScroll` are private Avalonia internals accessed via reflection. If Avalonia renames or removes them in a future version, this breaks silently — the scrollbar `Value` would update but the grid content wouldn't visually scroll. The null checks degrade gracefully, but worth calling out as a maintenance risk.

**2. Scroll direction may be inverted**

The delta uses `scrollStartH + delta.X`, which moves the scrollbar in the same direction as the mouse. Typical pan behavior drags the *content*, not the scrollbar thumb — dragging right should reveal content to the left. You might want to negate the deltas:

```csharp
hBar.Value = Math.Clamp(scrollStartH - delta.X, hBar.Minimum, hBar.Maximum);
// ...
vBar.Value = Math.Clamp(scrollStartV - delta.Y, vBar.Minimum, vBar.Maximum);
```

Test both directions and see which feels more natural. Browser and PDF viewer pan behavior uses the negated version.

**3. `Opened` change is unrelated**

The `MainWindow` change deferring plan loading to the `Opened` event is a separate fix. Not a blocker, but ideally would be its own commit so bisecting is easier if something regresses.

ClaudioESSilva:
@erikdarlingdata all (3) changes done.

erikdarlingdata (approved):
All three items addressed cleanly — reflection replaced with public Thumb.DragDeltaEvent, scroll direction fixed, and the Opened change separated out. Thanks for the contribution and for the quick turnaround on the feedback, @ClaudioESSilva!

PR #236, opened 2026-04-17
Joe feedback b1: Rule 34 — bare scan with narrow output

erikdarlingdata:
## Summary
New Rule 34 addressing Joe's b1 feedback: when a Clustered Index Scan or heap Table Scan has no predicate and outputs only a few columns, suggest adding a nonclustered index that could cover the query with less I/O (or columnstore for analytical workloads).

### Criteria
- `PhysicalOp` is "Clustered Index Scan" or "Table Scan"
- No predicate
- Output column count ≤ 3
- Actual plans: operator has non-zero own elapsed time
- Estimated plans: operator `CostPercent >= 20`

### Benefit scoring
Added "Bare Scan" to `OperatorTimeRules` — benefit = operator's own elapsed time / statement elapsed, same pattern as other scan rules.

## Test plan
- [x] 72 tests pass
- [ ] Load Joe's plan (ALa237EaOO) and verify the clustered index scan now gets a benefit % in the warnings list
- [ ] Verify the rule does NOT fire on scans with predicates (Rule 11 territory) or wide output lists

### DIFF
--- a/src/PlanViewer.Core/Services/PlanAnalyzer.cs
+++ b/src/PlanViewer.Core/Services/PlanAnalyzer.cs
@@ -892,6 +892,40 @@ public static class PlanAnalyzer
             }
         }
 
+        // Rule 34: Bare scan with narrow output — NC index or columnstore candidate.
+        // When a Clustered Index Scan or heap Table Scan reads the full table with no
+        // predicate but only outputs a few columns, a narrower nonclustered index could
+        // cover the query with far less I/O. For analytical workloads, columnstore may
+        // be a better fit.
+        var isBareScanCandidate = (node.PhysicalOp == "Clustered Index Scan" || node.PhysicalOp == "Table Scan")
+            && !node.Lookup
+            && string.IsNullOrEmpty(node.Predicate)
+            && !string.IsNullOrEmpty(node.OutputColumns);
+        if (!cfg.IsRuleDisabled(34) && isBareScanCandidate)
+        {
+            var colCount = node.OutputColumns!.Split(',').Length;
+            var isSignificant = node.HasActualStats
+                ? GetOperatorOwnElapsedMs(node) > 0
+                : node.CostPercent >= 20;
+
+            if (colCount <= 3 && isSignificant)
+            {
+                var scanKind = node.PhysicalOp == "Clustered Index Scan"
+                    ? "Clustered index scan"
+                    : "Heap table scan";
+                var indexAdvice = node.PhysicalOp == "Clustered Index Scan"
+                    ? "Consider a nonclustered index on the output columns (as key or INCLUDE) so SQL Server can read a narrower structure."
+                    : "Consider a clustered or nonclustered index on the output columns so SQL Server can read a narrower structure.";
+
+                node.Warnings.Add(new PlanWarning
+                {
+                    WarningType = "Bare Scan",
+                    Message = $"{scanKind} reads the full table with no predicate, outputting {colCount} column(s): {Truncate(node.OutputColumns, 200)}. {indexAdvice} For analytical workloads, a columnstore index may be a better fit.",
+                    Severity = PlanWarningSeverity.Warning
+                });
+            }
+        }
+
         // Rule 13: Mismatched data types (GetRangeWithMismatchedTypes / GetRangeThroughConvert)
         if (!cfg.IsRuleDisabled(13) && node.PhysicalOp == "Compute Scalar" && !string.IsNullOrEmpty(node.DefinedValues))
         {
--- a/src/PlanViewer.Core/Services/BenefitScorer.cs
+++ b/src/PlanViewer.Core/Services/BenefitScorer.cs
@@ -22,6 +22,7 @@ public static class BenefitScorer
         "Scan With Predicate",  // Rule 11
         "Non-SARGable Predicate", // Rule 12
         "Scan Cardinality Misestimate", // Rule 32
+        "Bare Scan",            // Rule 34
     };
 
     public static void Score(ParsedPlan plan)


src/PlanViewer.Core/Services/PlanAnalyzer.cs
```diff
@@ -892,6 +892,40 @@ _ when nonSargableReason.StartsWith("Function call") =>
...
         }
 
+        // Rule 34: Bare scan with narrow output — NC index or columnstore candidate.
+        // When a Clustered Index Scan or heap Table Scan reads the full table with no
+        // predicate but only outputs a few columns, a narrower nonclustered index could
+        // cover the query with far less I/O. For analytical workloads, columnstore may
+        // be a better fit.
+        var isBareScanCandidate = (node.PhysicalOp == "Clustered Index Scan" || node.PhysicalOp == "Table Scan")
+            && !node.Lookup
+            && string.IsNullOrEmpty(node.Predicate)
```
erikdarlingdata, line 902: Rules 11, 32, and 33 all gate on `IsRowstoreScan(node)`, which also admits non-clustered `Index Scan`. This rule instead hard-codes `PhysicalOp == "Clustered Index Scan" || "Table Scan"`. If the intent is to exclude NC index scans (because the advice "add a NC index" doesn't apply), that's worth a one-line comment — otherwise consider `IsRowstoreScan(node) && node.PhysicalOp != "Index Scan"` or similar, to stay consistent with the other scan rules and avoid a future drift point.

---

src/PlanViewer.Core/Services/PlanAnalyzer.cs
```diff
@@ -892,6 +892,40 @@ _ when nonSargableReason.StartsWith("Function call") =>
...
+            && !string.IsNullOrEmpty(node.OutputColumns);
+        if (!cfg.IsRuleDisabled(34) && isBareScanCandidate)
+        {
+            var colCount = node.OutputColumns!.Split(',').Length;
+            var isSignificant = node.HasActualStats
+                ? GetOperatorOwnElapsedMs(node) > 0
+                : node.CostPercent >= 20;
```
erikdarlingdata, line 909: Significance thresholds are much looser and asymmetric compared to sibling scan rules:

- Actual path: `GetOperatorOwnElapsedMs(node) > 0` — any scan that registers ≥1ms of own time fires. Rule 11 elevates severity at `CostPct >= 90 / ElapsedPct >= 90`; Rule 32 gates at `>= 50`. A 1ms threshold will fire on nearly every non-trivial OLTP plan with a bare scan — expect noise.
- Estimated path: `CostPercent >= 20` — Rule 33 uses `>= 50` for the analogous case. Worth aligning (or documenting why 20% is correct here).

Consider reusing `BuildScanImpactDetails(node, stmt)` and a `CostPct/ElapsedPct >= 50` gate like the other scan rules, so the rule set has one consistent definition of "significant scan."

---

src/PlanViewer.Core/Services/PlanAnalyzer.cs
```diff
@@ -892,6 +892,40 @@ _ when nonSargableReason.StartsWith("Function call") =>
...
+
+            if (colCount <= 3 && isSignificant)
```
erikdarlingdata, line 911: `colCount <= 3` is a magic number — fine as a heuristic but hoist it to a named local/const (e.g. `NarrowOutputMaxColumns = 3`) so future tuning is obvious. Also note: `OutputColumns` is joined with `", "` by `ShowPlanParser` (line 1214) and brackets are stripped, so `Split(',').Length` is safe — no risk of embedded commas breaking the count.

---

src/PlanViewer.Core/Services/PlanAnalyzer.cs
```diff
@@ -892,6 +892,40 @@ _ when nonSargableReason.StartsWith("Function call") =>
...
+                    ? "Clustered index scan"
+                    : "Heap table scan";
+                var indexAdvice = node.PhysicalOp == "Clustered Index Scan"
+                    ? "Consider a nonclustered index on the output columns (as key or INCLUDE) so SQL Server can read a narrower structure."
+                    : "Consider a clustered or nonclustered index on the output columns so SQL Server can read a narrower structure.";
+
+                node.Warnings.Add(new PlanWarning
+                {
+                    WarningType = "Bare Scan",
+                    Message = $"{scanKind} reads the full table with no predicate, outputting {colCount} column(s): {Truncate(node.OutputColumns, 200)}. {indexAdvice} For analytical workloads, a columnstore index may be a better fit.",
```
erikdarlingdata, line 923: No operator-time / cost context is surfaced in the message. Every other scan rule (11, 32, 33) appends "Operator time: Xms (Y% of statement)" or equivalent via `BuildScanImpactDetails` so the reader can gauge severity. Consider adding the same here — right now the warning gives no signal to the user about whether this bare scan is 1ms or 10 minutes.

---

erikdarlingdata (commented):
## Summary
Adds Rule 34 "Bare Scan": fires on Clustered Index Scan or heap Table Scan with no predicate and ≤3 output columns, suggesting a narrower NC index (or columnstore for analytical workloads). Registered in `BenefitScorer.OperatorTimeRules` so the benefit % is computed the same way as other scan rules.

## Good
- Base branch is `dev` ✓
- Gated on `cfg.IsRuleDisabled(34)` and `!node.Lookup`, so Rule 10 (key/RID lookups) won't double-fire.
- Correct split safety — `ShowPlanParser.cs:1214` joins with `", "` and strips brackets, so `Split(',').Length` is accurate.
- `BenefitScorer` entry mirrors the other scan rules' pattern.

## Needs attention

1. **Cross-repo sync (required).** This changes the PlanAnalyzer rule set. The equivalent rule must also land in PerformanceMonitor's Dashboard and Lite repos — they carry their own copy of the rule set and are expected to stay in sync with Studio. Confirm that follow-up PRs are tracked.

2. **No test coverage.** `tests/PlanViewer.Core.Tests/PlanAnalyzerTests.cs` has no new test for Rule 34 — the PR body's own test plan lists two unchecked manual items (Joe's plan `ALa237EaOO`, and negative cases for scans with predicates / wide outputs). At minimum, add a fixture-based test asserting: (a) fires on bare CIX scan with ≤3 cols, (b) does not fire when a predicate exists, (c) does not fire when output is wide.

3. **Threshold inconsistency with sibling scan rules** (see inline on `PlanAnalyzer.cs:907-909`). Actual-plan significance is `own elapsed > 0ms`; Rule 11 / 32 use `>= 50%` cost or elapsed. Estimated path is `CostPercent >= 20`; Rule 33 uses `>= 50`. Expect this rule to be noisier than its neighbors unless tightened.

4. **Does not use `IsRowstoreScan` helper** (inline on `PlanAnalyzer.cs:902`). If excluding NC `Index Scan` is intentional, add a one-line comment; otherwise consider routing through the same helper the other scan rules use.

5. **Warning message lacks impact context** (inline on `PlanAnalyzer.cs:923`). No operator-time / cost figure is surfaced, unlike every other scan rule. Users have nothing to gauge severity by.

No Avalonia/XAML surface in this PR, so those review items are N/A. Comment-only review — leaving the call to the maintainer.

---

PR #241, opened 2026-04-20
Await ShowDialog in Format_Click (fix CS4014)

erikdarlingdata:
## Summary
- Awaits `dialog.ShowDialog(GetParentWindow())` in `Format_Click` so the status update and method return wait for the user to dismiss the parse-error dialog
- Eliminates the new CS4014 warning introduced by #238

## Test plan
- [x] `dotnet build -c Debug` — 0 warnings, 0 errors
- [x] App launches and runs

### DIFF
--- a/src/PlanViewer.App/Controls/QuerySessionControl.axaml.cs
+++ b/src/PlanViewer.App/Controls/QuerySessionControl.axaml.cs
@@ -2056,7 +2056,7 @@ public partial class QuerySessionControl : UserControl
                         }
                     }
                 };
-                dialog.ShowDialog(GetParentWindow());
+                await dialog.ShowDialog(GetParentWindow());
                 SetStatus($"Format failed: {errors.Count} error(s)");
                 return;
             }


erikdarlingdata:
## Review

**What it does:** Adds `await` to `dialog.ShowDialog(GetParentWindow())` in `Format_Click` (src/PlanViewer.App/Controls/QuerySessionControl.axaml.cs:2059) to clear the CS4014 warning introduced by #238.

**Good:**
- Base branch is `dev` ✓
- Minimal, correctly scoped fix. The enclosing method is already `async void`, so `await` is the right call.
- Behavior is now more correct: `SetStatus("Format failed: …")` and the `finally { FormatButton.IsEnabled = true; }` now run after the user dismisses the modal, not while it's still open. That's the intended UX — the button shouldn't re-enable behind a live error dialog.
- No Avalonia gotchas, no brush/contrast changes, no PlanAnalyzer rule changes, no MVVM creep.

**Notes:**
- `FormatOptions_Click` at the bottom of the same file (line ~2089) uses the same fire-and-forget `dialog.ShowDialog(GetParentWindow())` pattern, but in a non-async `void` handler with nothing after it, so no CS4014 and no behavior bug. Out of scope here; just flagging for awareness if the same treatment is ever desired for consistency.
- No test added, but the change is a pure UI interaction ordering fix in `PlanViewer.App` (not `PlanViewer.Core`), and `tests/PlanViewer.Core.Tests/` wouldn't be the right home for it. Acceptable.

No blockers from my end.

---

PR #234, opened 2026-04-17
Fix: Query Editor losing syntax highlighting on tab switch

ClaudioESSilva:
## What does this PR do?

Fixes #233 

## Which component(s) does this affect?

- [x] Desktop App (PlanViewer.App)
- [ ] Core Library (PlanViewer.Core)
- [ ] CLI Tool (PlanViewer.Cli)
- [ ] SSMS Extension (PlanViewer.Ssms)
- [ ] Tests
- [ ] Documentation

## How was this tested?

https://github.com/user-attachments/assets/dd4296c4-05d9-4c96-8747-982535d4fd25

## Checklist

- [x] I have read the [contributing guide](https://github.com/erikdarlingdata/PerformanceStudio/blob/main/CONTRIBUTING.md)
- [X] My code builds with zero warnings (`dotnet build -c Debug`)
- [x] All tests pass (`dotnet test`)
- [x] I have not introduced any hardcoded credentials or server names

PR #218, opened 2026-04-12
Feature/query store group by changer

rferraton:
## What does this PR do?

In the Query Store grid, Add a **Group By** Selector and the capacity to group (and search) for Top Query (using Query Hash) and Top Module. It build a hierachy from the Group By selection to the QueryId/PlanId.
There is a **Group By** 
- **None** that let the grid as is.
- **Query Hash** that build a hierarchy with QueryHash > PlanHash > Top/Bottom QueryID+PlanId
- **Module** that build a hierarchy with Module> QueryHash > Top/Bottom QueryID+PlanId

## Which component(s) does this affect?

- [x] Desktop App (PlanViewer.App)
- [x] Core Library (PlanViewer.Core)
- [ ] CLI Tool (PlanViewer.Cli)
- [ ] SSMS Extension (PlanViewer.Ssms)
- [ ] Tests
- [ ] Documentation

## How was this tested?
2026-04-12_17h22_21

Describe the testing you've done. Include:
- Plan files tested : Query Store, Load Plan from query Store
- Platforms tested : Windows Only

## Checklist

- [x] I have read the [contributing guide](https://github.com/erikdarlingdata/PerformanceStudio/blob/main/CONTRIBUTING.md)
- [x] My code builds with zero warnings (`dotnet build -c Debug`)
- [x] All tests pass (`dotnet test`)
- [x] I have not introduced any hardcoded credentials or server names

PR #243, opened 2026-04-20
Release v1.6.0

erikdarlingdata:
### Highlights
- **Wait stats benefit scoring (Stage 2 of #215)** — wait stats now contribute to the max-benefit calculation in PlanAnalyzer findings
- **CPU:Elapsed ratio** replaces the old parallelism-efficiency warnings for clearer parallel-plan diagnosis
- **SQL Formatter** in the Query Editor (#238) with persistent format options (parameters dialog, async formatting, parse-error popup)
- **Query Store group-by changer** (#218) — group plans by QueryHash or Module, with auto-expand, golden headers, and reordered columns
- **Built-in analytics** for PlanShare and stats dashboard
- **Joe's plan-review feedback** — false-positive fixes, improved messaging, Rule 34 narrowing, Serial Plan / adaptive join math, MAXDOP 1 wording, Scalar UDF redundancy, parallel-skew + key-lookup NL join scoring

### Other
- PlanViewer toolbars / Query Store headers / tab accent polish
- Query Editor no longer loses syntax highlighting on tab switch (#234)
- Auto-scrolling improvements (#225, thanks @ClaudioESSilva)
- Wait-stats benefit wired into the web plan viewer

### DIFF
--- a/src/PlanViewer.App/Controls/QuerySessionControl.axaml.cs
+++ b/src/PlanViewer.App/Controls/QuerySessionControl.axaml.cs
@@ -2056,7 +2056,7 @@ public partial class QuerySessionControl : UserControl
                         }
                     }
                 };
-                dialog.ShowDialog(GetParentWindow());
+                await dialog.ShowDialog(GetParentWindow());
                 SetStatus($"Format failed: {errors.Count} error(s)");
                 return;
             }
--- a/src/PlanViewer.App/Controls/QuerySessionControl.axaml.cs
+++ b/src/PlanViewer.App/Controls/QuerySessionControl.axaml.cs
@@ -2005,4 +2005,86 @@ public partial class QuerySessionControl : UserControl
         var parent = this.VisualRoot;
         return parent as Window ?? throw new InvalidOperationException("No parent window");
     }
+
+    private async void Format_Click(object? sender, RoutedEventArgs e)
+    {
+        var sql = QueryEditor.Text;
+        if (string.IsNullOrWhiteSpace(sql))
+            return;
+
+        FormatButton.IsEnabled = false;
+        SetStatus("Formatting...");
+
+        try
+        {
+            var settings = SqlFormatSettingsService.Load(out var loadError);
+            if (loadError != null)
+                SetStatus("Warning: using default format settings (load failed)");
+
+            var (formatted, errors) = await Task.Run(() => SqlFormattingService.Format(sql, settings));
+
+            if (errors != null && errors.Count > 0)
+            {
+                var errorMessages = string.Join("\n", errors.Select(err => $"Line {err.Line}: {err.Message}"));
+                var dialog = new Window
+                {
+                    Title = "SQL Format Error",
+                    Width = 500,
+                    Height = 250,
+                    WindowStartupLocation = WindowStartupLocation.CenterOwner,
+                    Icon = GetParentWindow().Icon,
+                    Background = (IBrush)this.FindResource("BackgroundBrush")!,
+                    Foreground = (IBrush)this.FindResource("ForegroundBrush")!,
+                    Content = new StackPanel
+                    {
+                        Margin = new Avalonia.Thickness(20),
+                        Children =
+                        {
+                            new TextBlock
+                            {
+                                Text = $"Could not format: {errors.Count} parse error(s)",
+                                FontWeight = Avalonia.Media.FontWeight.Bold,
+                                FontSize = 14,
+                                Margin = new Avalonia.Thickness(0, 0, 0, 10)
+                            },
+                            new TextBlock
+                            {
+                                Text = errorMessages,
+                                TextWrapping = TextWrapping.Wrap,
+                                FontSize = 12
+                            }
+                        }
+                    }
+                };
+                dialog.ShowDialog(GetParentWindow());
+                SetStatus($"Format failed: {errors.Count} error(s)");
+                return;
+            }
+
+            var caretOffset = QueryEditor.CaretOffset;
+
+            QueryEditor.Document.BeginUpdate();
+            try
+            {
+                QueryEditor.Document.Replace(0, QueryEditor.Document.TextLength, formatted);
+            }
+            finally
+            {
+                QueryEditor.Document.EndUpdate();
+            }
+
+            QueryEditor.CaretOffset = Math.Min(caretOffset, QueryEditor.Document.TextLength);
+            SetStatus("Formatted");
+        }
+        finally
+        {
+            FormatButton.IsEnabled = true;
+        }
+    }
+
+    private void FormatOptions_Click(object? sender, RoutedEventArgs e)
+    {
+        var dialog = new Dialogs.FormatOptionsWindow();
+        dialog.ShowDialog(GetParentWindow());
+    }
 }
--- a/src/PlanViewer.App/Controls/QuerySessionControl.axaml.cs
+++ b/src/PlanViewer.App/Controls/QuerySessionControl.axaml.cs
@@ -68,13 +68,22 @@ public partial class QuerySessionControl : UserControl
         QueryEditor.TextArea.TextEntered += OnTextEntered;
 
         // Focus the editor when the control is attached to the visual tree
+        // Re-install TextMate if it was disposed on detach (tab switching disposes it)
         AttachedToVisualTree += (_, _) =>
         {
+            if (_textMateInstallation == null)
+                SetupSyntaxHighlighting();
+
             QueryEditor.Focus();
             QueryEditor.TextArea.Focus();
         };
 
-        DetachedFromVisualTree += (_, _) => _textMateInstallation?.Dispose();
+        // Dispose TextMate when detached (e.g. tab switch) to release renderers/transformers
+        DetachedFromVisualTree += (_, _) =>
+        {
+            _textMateInstallation?.Dispose();
+            _textMateInstallation = null;
+        };
 
         // Focus the editor when the Editor tab is selected; toggle plan-dependent buttons
         SubTabControl.SelectionChanged += (_, _) =>
--- a/src/PlanViewer.App/Dialogs/FormatOptionsWindow.axaml.cs
+++ b/src/PlanViewer.App/Dialogs/FormatOptionsWindow.axaml.cs
@@ -1,0 +1,326 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.ComponentModel;
+using System.Diagnostics;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+using Avalonia.Controls;
+using Avalonia.Interactivity;
+using Avalonia.Media;
+using PlanViewer.App.Services;
+
+namespace PlanViewer.App.Dialogs;
+
+public partial class FormatOptionsWindow : Window
+{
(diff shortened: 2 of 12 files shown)


src/PlanViewer.App/Dialogs/FormatOptionsWindow.axaml.cs
```diff
@@ -0,0 +1,326 @@
...
+        discardBtn.Click += (_, _) => { tcs.TrySetResult(true); dialog.Close(); };
+        cancelBtn.Click += (_, _) => { tcs.TrySetResult(false); dialog.Close(); };
+        dialog.Closing += (_, _) => tcs.TrySetResult(false);
+
+        await dialog.ShowDialog(this);
+        return await tcs.Task;
+    }
+}
+
+public class FormatOptionRow : INotifyPropertyChanged
```
erikdarlingdata, line 278: `FormatOptionRow : INotifyPropertyChanged` is MVVM creep. Repo convention is code-behind only — no INotifyPropertyChanged, no bindings-to-VM. The DataGrid needs change notification only for the `BoolValue`⇄`CurrentValue` sync; that can be handled imperatively from the code-behind (e.g., wire `ToggleSwitch.IsCheckedChanged` in cell templates, or rebuild the rows after edits). If you keep INPC here, it sets a precedent that will leak into the rest of the app.

---

src/PlanViewer.App/Dialogs/FormatOptionsWindow.axaml.cs
```diff
@@ -0,0 +1,326 @@
...
+                }
+                else
+                    value = row.CurrentValue;
+
+                prop.SetValue(settings, value);
+            }
+            catch (Exception ex)
+            {
+                Debug.WriteLine($"FormatOptions: failed to set {row.Name}: {ex.Message}");
+            }
```
erikdarlingdata, line 124: Both error paths here silently drop the user's edit with only a `Debug.WriteLine`:

- L109-113: invalid int → `continue`, the previously-saved value is silently kept.
- L121-124: blanket `catch (Exception)` → same.

From the user's perspective, they typed something, clicked Save, the dialog closed, and the setting didn't change. There's no feedback. Either surface a validation error (reuse `ShowErrorPopup` with `return;` so nothing saves) or constrain the editor so a bad value can't be typed in the first place. Also consider `int.TryParse(row.CurrentValue, NumberStyles.Integer, CultureInfo.InvariantCulture, out …)` since the JSON file is invariant.

---

src/PlanViewer.App/Dialogs/FormatOptionsWindow.axaml.cs
```diff
@@ -0,0 +1,326 @@
...
+        if (_isDirty)
+        {
+            e.Cancel = true;
+            TryClose();
+            return;
+        }
+        base.OnClosing(e);
+    }
+
+    private async void TryClose()
```
erikdarlingdata, line 189: `OnClosing` fires-and-forgets an `async void TryClose()`. Any exception from `ShowDiscardDialog` (e.g., `FindResource` miss on a themed resource, dialog owner already closed) propagates to the `SynchronizationContext` as an unhandled exception and crashes the app. Wrap the body of `TryClose` in `try/catch` and log, or make it return `Task` and await it from a single-purpose `async void` shim. Same concern applies to `Format_Click` in `QuerySessionControl.axaml.cs:2009` — the `try/finally` there doesn't catch, so `SqlFormattingService.Format` / `ShowDialog` exceptions bubble out of an `async void`.

---

src/PlanViewer.App/Dialogs/FormatOptionsWindow.axaml.cs
```diff
@@ -0,0 +1,326 @@
...
+
+    private void Revert_Click(object? sender, RoutedEventArgs e)
+    {
+        foreach (var row in _rows)
+        {
+            row.CurrentValue = row.DefaultValue;
+            if (row.IsBool)
+                row.BoolValue = row.DefaultBoolValue;
+        }
+    }
```
erikdarlingdata, line 144: `Revert_Click` mutates `CurrentValue` / `BoolValue`, which fires the `PropertyChanged` handler added in `LoadSettings` (L91) and flips `_isDirty = true`, even when the user reverts to defaults and never had pending edits. Net effect: click Revert → Close → "Unsaved Changes" prompt appears unnecessarily. Consider suppressing the dirty flag during revert, or comparing against defaults before setting `_isDirty`.

---

src/PlanViewer.App/Services/SqlFormattingService.cs
```diff
@@ -0,0 +1,72 @@
...
+        settings ??= new SqlFormatSettings();
+
+        var parser = GetParser(settings.SqlVersion);
+
+        using var reader = new StringReader(sql);
+        var fragment = parser.Parse(reader, out var errors);
+
+        if (fragment == null)
+            return (sql, errors);
+
```
erikdarlingdata, line 28: If `parser.Parse` returns a non-null fragment but `errors` is non-empty (it's designed to recover), you still run `GenerateScript` against the partially-valid AST and return it as `formattedText`. Callers in `QuerySessionControl.Format_Click` (L2026) treat any non-null `errors` as "format failed, show dialog and bail" — so the generated output is discarded. That's the safer path, but it means this method is doing work that gets thrown away. Either bail here too when errors are present, or let callers decide without double-generating.

---

src/PlanViewer.App/Controls/QuerySessionControl.axaml.cs
```diff
@@ -68,13 +68,22 @@ public QuerySessionControl(ICredentialService credentialService, ConnectionStore
...
             QueryEditor.TextArea.Focus();
         };
 
-        DetachedFromVisualTree += (_, _) => _textMateInstallation?.Dispose();
+        // Dispose TextMate when detached (e.g. tab switch) to release renderers/transformers
+        DetachedFromVisualTree += (_, _) =>
+        {
+            _textMateInstallation?.Dispose();
+            _textMateInstallation = null;
+        };
```
erikdarlingdata, line 86: Fix for #234 looks right. Two notes:

1. `SetupSyntaxHighlighting()` is called on every re-attach; confirm it doesn't leak the `RegistryOptions` / grammar objects (the old `_textMateInstallation?.Dispose()` only releases the installation, not whatever `InstallTextMate` allocated on the editor). If it's cheap, fine.
2. Tab-switch churn now runs a full TextMate reinstall every time — worth a quick perf sanity check with a busy session that has many tabs.

---

src/PlanViewer.App/Services/SqlFormatSettingsService.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            AlignClauseBodies = AlignClauseBodies,
+            AlignColumnDefinitionFields = AlignColumnDefinitionFields,
+            AlignSetClauseItem = AlignSetClauseItem,
+            AsKeywordOnOwnLine = AsKeywordOnOwnLine,
+            IncludeSemicolons = IncludeSemicolons,
+            IndentSetClause = IndentSetClause,
+            IndentViewBody = IndentViewBody,
+            IndentationSize = IndentationSize,
+            KeywordCasing = Enum.TryParse<Microsoft.SqlServer.TransactSql.ScriptDom.KeywordCasing>(KeywordCasing, true, out var kc)
```
erikdarlingdata, line 58: `KeywordCasing` round-trips via `ToString()` / `Enum.TryParse`. Persisting the string form is fine, but there's no validation on Save — a user can type any text in the options grid (if the combo cell is edited as text), it'll get saved to JSON, and on next Load the `Enum.TryParse` silently falls back to `Uppercase`. Not a real risk with the current ComboBox-backed editor, but worth validating on `Save` so bad state never hits disk.

---

erikdarlingdata (commented):
## PR #243 — v1.6.0 release review

**What it does**
Merges `dev` → `main` for v1.6.0. The 12 files in this diff cover: the new SQL Formatter (`SqlFormattingService` + `SqlFormatSettingsService` + `FormatOptionsWindow` + Format/Format Options buttons in `QuerySessionControl`), a fix for Query Editor losing syntax highlighting on tab switch (#234), PlanViewer / Query Store toolbar polish (padding/spacing/column width tweaks and header renames), two WaitCategory color swaps in `DarkTheme.axaml`, a Fluent `SystemAccentColor*` override, a removed unused `using System;` in `ServerConnection.cs`, the v1.6.0 csproj bump, and `THIRD-PARTY-NOTICES.md` for SqlFormatter attribution.

**Base branch: `main`.** Per the standard rule, PRs target `dev` — but this is the release merge, so it's intentional. Calling it out for the record.

**PR description vs. diff mismatch.** The description claims wait-stats benefit scoring (Stage 2 of #215), CPU:Elapsed ratio, Query Store group-by changer (#218), PlanShare analytics, Joe's plan-review feedback (Rule 34 narrowing, Serial Plan / adaptive join math, Scalar UDF, parallel-skew + key-lookup NL scoring), etc. None of those are in this diff — there are no `PlanViewer.Core` rule/scoring changes here. Presumably they landed via earlier dev→main merges. Worth reconciling the release notes before tagging, otherwise the changelog will mislead users scanning the commit range.

**PlanAnalyzer sync check**: Since no PlanAnalyzer rules are touched here, nothing to sync to PerformanceMonitor Dashboard / Lite for this PR. If any rule changes did land on `dev` since the last release, verify those were mirrored at the time.

**What's good**
- SQL Formatter is correctly implemented as a `Task.Run` wrapper so parsing doesn't block the UI thread.
- `QueryEditor.Document.BeginUpdate()` / `EndUpdate()` around `Replace` preserves undo and avoids intermediate re-rendering; caret offset is clamped to new length.
- `SetupSyntaxHighlighting` re-install on re-attach is a clean fix for the #234 tab-switch issue.
- `PlaceholderText` (not `Watermark`) is used correctly on ComboBoxes.
- New `WaitCategory` colors (`#52E3B5`, `#7B4FFF`) are bright/saturated — not in the rejected `#6B7280`/dim-grey range.
- `SqlFormatSettingsService` round-trips to LocalApplicationData with reasonable error surfaces.

**What needs attention** (see inline comments)
1. `FormatOptionRow : INotifyPropertyChanged` — MVVM creep that the repo explicitly avoids.
2. `Save_Click` silently swallows bad int input and generic exceptions; the user gets no feedback that their edit was dropped.
3. `OnClosing` fire-and-forgets an `async void TryClose()` that can crash the app on exception. `Format_Click` has the same `async void` exposure.
4. `Revert_Click` flips `_isDirty=true` unnecessarily, triggering a spurious "Unsaved Changes" prompt.
5. `SqlFormattingService.Format` generates the script even when parse errors exist and callers discard it.
6. `SetupSyntaxHighlighting` re-install — quick perf/leak sanity check on heavy multi-tab sessions.
7. No validation on `KeywordCasing` string at save time (low risk given the ComboBox UI).

**Test coverage.** New logic in `src/PlanViewer.App/Services/` (SqlFormattingService, SqlFormatSettingsService) has no corresponding tests. `tests/PlanViewer.Core.Tests/` is `Core`-scoped so it isn't the natural home — but if there's no `PlanViewer.App.Tests`, the settings round-trip and format-happy-path at minimum are easy to cover and would catch regressions on the ScriptDom dependency bump.

Comments only — not approving or requesting changes. Maintainer's call.

---

PR #244, opened 2026-04-21
Wait stats as warnings (#215, Joe feedback a1/b4/b5)

erikdarlingdata:
## Summary
- Each recorded wait stat emits a first-class `PlanWarning` with per-wait description, how-to-fix guidance, and the benefit % already calculated by `BenefitScorer`. Joe's a1/b4/b5 asks from #215 — integrate waits into the warnings list instead of leaving them in their own card.
- Wait-stats chart stays as a complementary view (editorial call from the issue thread).

### New: `WaitStatsKnowledge.cs`
Static content library mapping wait type → description + how-to-fix. Covers:
- I/O: `PAGEIOLATCH_SH/EX/UP/DT`, `WRITELOG`, `IO_COMPLETION`, `ASYNC_IO_COMPLETION`, `LOGBUFFER`
- Memory: `MEMORY_ALLOCATION_EXT`, `RESOURCE_SEMAPHORE`, `RESOURCE_SEMAPHORE_QUERY_COMPILE`, `SOS_PHYS_PAGE_CACHE`
- Scheduler: `SOS_SCHEDULER_YIELD`, `THREADPOOL`, `DISPATCHER_QUEUE_SEMAPHORE`
- Parallelism: `CXPACKET`, `CXCONSUMER`, `CXSYNC_PORT`, `CXSYNC_CONSUMER`
- Batch-mode hash: `HTBUILD`, `HTREPARTITION`, `HTDELETE`, `HTMEMO`, `HTREINIT`, `BPSORT`, `BMPBUILD`
- Latch: `PAGELATCH_SH/EX/UP`, `LATCH_EX`, `LATCH_SH`
- Lock: `LCK_M_S/X/U/IS/IX/SCH_S/SCH_M/RS_S/RS_U/RX_X`
- Network: `ASYNC_NETWORK_IO`
- Misc: `EXECSYNC`, `PREEMPTIVE_OS_WRITEFILEGATHER`

Prefix fallbacks for `LCK_M_`, `PAGEIOLATCH_`, `PAGELATCH_`, `LATCH_`, `HT`, `CX`, `PREEMPTIVE_`, plus a `MEMORY_ALLOCATION` contains-match (catches `RESERVED_MEMORY_ALLOCATION_EXT`).

### `BenefitScorer.EmitWaitStatWarnings`
Runs after `ScoreWaitStats`. For each `WaitStatInfo` it builds a warning:
- `WarningType = "Wait: {WaitType}"`
[...]

PR #245, opened 2026-04-21
Release v1.7.0 — wait stats as warnings

erikdarlingdata:
## Summary
- Single PR merging PR #244: **wait stats as warnings** (#215, Joe feedback a1/b4/b5). Each wait recorded on a statement now emits a first-class warning with per-wait description, how-to-fix guidance, and benefit %.
- `PAGEIOLATCH_*` warnings include effective latency per wait (wait_ms / wait_count).
- Wait-stats chart stays as a complementary view.
- Flows through existing `PlanWarnings` pipeline → web viewer, HTML export, JSON/text, Avalonia Plan Warnings expander all updated with no renderer changes.

See PR #244 for full details.

## Test plan
- [x] CLI smoke on `20260415_1.sqlplan` and `NXfso4jC0L.sqlplan`
- [x] build-and-test passed on PR #244
- [ ] Web viewer visual check post-deploy
- [ ] Desktop Plan Warnings expander visual check post-deploy

### DIFF
--- a/src/PlanViewer.Core/Services/BenefitScorer.cs
+++ b/src/PlanViewer.Core/Services/BenefitScorer.cs
@@ -38,8 +38,71 @@ public static class BenefitScorer
 
                 if (stmt.WaitStats.Count > 0 && stmt.QueryTimeStats != null)
                     ScoreWaitStats(stmt);
+
+                if (stmt.WaitStats.Count > 0)
+                    EmitWaitStatWarnings(stmt);
             }
         }
+    }
+
+    /// <summary>
+    /// Emits a PlanWarning per wait stat entry, merging the per-wait benefit %
+    /// from ScoreWaitStats with the descriptive content from WaitStatsKnowledge.
+    /// The existing wait-stats chart/card stays as a complementary view.
+    /// </summary>
+    private static void EmitWaitStatWarnings(PlanStatement stmt)
+    {
+        // Lookup benefit % by wait type (populated by ScoreWaitStats)
+        var benefitByType = new Dictionary<string, double>(StringComparer.OrdinalIgnoreCase);
+        foreach (var wb in stmt.WaitBenefits)
+            benefitByType[wb.WaitType] = wb.MaxBenefitPercent;
+
+        foreach (var wait in stmt.WaitStats)
+        {
+            if (wait.WaitTimeMs <= 0) continue;
+
+            var entry = WaitStatsKnowledge.Lookup(wait.WaitType);
+            double? benefitPct = benefitByType.TryGetValue(wait.WaitType, out var b) ? b : null;
+
+            var msg = new System.Text.StringBuilder();
+            msg.Append(wait.WaitType).Append(": ").Append(entry.Description);
+            msg.Append(" Observed ").Append(wait.WaitTimeMs.ToString("N0")).Append(" ms");
+            if (wait.WaitCount > 0)
+                msg.Append(" across ").Append(wait.WaitCount.ToString("N0")).Append(" wait").Append(wait.WaitCount == 1 ? "" : "s");
+            msg.Append('.');
+
+            if (entry.ShowEffectiveLatency && wait.WaitCount > 0)
+            {
+                var effLatency = (double)wait.WaitTimeMs / wait.WaitCount;
+                msg.Append(" Effective latency: ")
+                   .Append(FormatLatency(effLatency))
+                   .Append(" per wait.");
+            }
+
+            var severity = benefitPct switch
+            {
+                >= 50 => PlanWarningSeverity.Critical,
+                >= 10 => PlanWarningSeverity.Warning,
+                _     => PlanWarningSeverity.Info,
+            };
+
+            stmt.PlanWarnings.Add(new PlanWarning
+            {
+                WarningType = "Wait: " + wait.WaitType,
+                Message = msg.ToString(),
+                Severity = severity,
+                MaxBenefitPercent = benefitPct,
+                ActionableFix = entry.HowToFix
+            });
+        }
+    }
+
+    private static string FormatLatency(double ms)
+    {
+        if (ms >= 1000) return $"{ms / 1000:N2} s";
+        if (ms >= 10)   return $"{ms:N0} ms";
+        if (ms >= 1)    return $"{ms:N1} ms";
+        return $"{ms * 1000:N0} µs";
     }
 
     private static void ScoreStatementWarnings(PlanStatement stmt)
--- a/src/PlanViewer.Core/Services/WaitStatsKnowledge.cs
+++ b/src/PlanViewer.Core/Services/WaitStatsKnowledge.cs
@@ -1,0 +1,307 @@
+using System;
+using System.Collections.Generic;
+
+namespace PlanViewer.Core.Services;
+
+/// <summary>
+/// Per-wait-type knowledge used when surfacing wait stats as warnings:
+/// what the wait means, how to address it, and any per-wait display hints.
+/// Entries are looked up by exact wait type first, then by prefix/family fallback,
+/// and finally a generic default.
+/// </summary>
+public static class WaitStatsKnowledge
+{
+    public sealed class Entry
+    {
+        /// <summary>Short, human-readable explanation of what the wait represents.</summary>
+        public string Description { get; init; } = "";
+
+        /// <summary>Concrete guidance on how to reduce or eliminate the wait.</summary>
+        public string HowToFix { get; init; } = "";
+
+        /// <summary>
+        /// If true, surface an effective per-wait latency (wait_ms / wait_count)
+        /// in the warning message. Useful for latch/I/O waits where tail latency is
+        /// the thing to focus on.
+        /// </summary>
+        public bool ShowEffectiveLatency { get; init; }
+    }
+
+    private static readonly Entry Default = new()
+    {
+        Description = "Query time was spent waiting on this resource.",
+        HowToFix = "Investigate why this wait is elevated for this query — the wait type name is the best starting point."
+    };
+
+    // Exact-match lookup. Prefix fallbacks handled in Lookup().
+    private static readonly Dictionary<string, Entry> Exact = new(StringComparer.OrdinalIgnoreCase)
+    {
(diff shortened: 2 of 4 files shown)


src/PlanViewer.Core/Services/WaitStatsKnowledge.cs
```diff
@@ -0,0 +1,307 @@
...
+
+        if (wt.StartsWith("PREEMPTIVE_"))
+            return new Entry
+            {
+                Description = "SQL Server is waiting on an external OS or system call.",
+                HowToFix = "The specific preemptive wait suffix indicates the source (file growth, DNS, CLR, etc.). Address whichever external dependency is slow."
+            };
+
+        if (wt.Contains("MEMORY_ALLOCATION"))
+            return Exact["MEMORY_ALLOCATION_EXT"];
```
erikdarlingdata, line 303: `Exact["MEMORY_ALLOCATION_EXT"]` is a direct dictionary access — if anyone later removes or renames that key in the `Exact` initializer above, this throws `KeyNotFoundException` at runtime instead of falling back to `Default`. Either `TryGetValue` it, or just inline a fresh `Entry { ... }` here so the fallback isn't coupled to the exact-match table.

---

src/PlanViewer.Core/Services/WaitStatsKnowledge.cs
```diff
@@ -0,0 +1,307 @@
...
+            };
+
+        if (wt.StartsWith("LATCH_"))
+            return new Entry
+            {
+                Description = "Waiting for a non-buffer latch on an internal SQL Server structure.",
+                HowToFix = "Check sys.dm_os_latch_stats for the specific latch class — these are usually concurrency signals rather than query-tuning problems."
+            };
+
+        if (wt.StartsWith("HT") && wt.Length > 2)
```
erikdarlingdata, line 281: `StartsWith("HT") && Length > 2` is too broad — it will swallow any future/unrelated wait type whose name happens to begin with `HT` and mislabel it as a batch-mode hash wait with batch-mode-hash how-to-fix advice. The known batch-mode hash family is `HTBUILD/HTREPARTITION/HTDELETE/HTMEMO/HTREINIT` (all already in `Exact`), so the prefix fallback adds no value today and risks wrong attribution tomorrow. Consider either dropping this branch entirely (let unknown HT* fall through to `Default`) or tightening to a known prefix like `HTREP` / explicit list.

---

src/PlanViewer.Core/Services/WaitStatsKnowledge.cs
```diff
@@ -0,0 +1,307 @@
...
+            return new Entry
+            {
+                Description = "Batch-mode hash operation synchronization wait.",
+                HowToFix = "Reduce input rows and skew for the batch-mode hash. Confirm the smaller side is on the build, and check that batch mode is the right choice for this plan shape."
+            };
+
+        if (wt.StartsWith("CX"))
```
erikdarlingdata, line 288: Same issue as the `HT` prefix above: `StartsWith("CX")` matches anything starting with `CX`. The full known parallel-exchange set (`CXPACKET/CXCONSUMER/CXSYNC_PORT/CXSYNC_CONSUMER`) is already in `Exact`. A bare `CX` prefix gives a generic "parallel exchange" answer for anything that *happens* to start with those two letters — same false-positive risk as `HT`. Either tighten to `CXSYNC_` / `CXPACKET` / `CXCONSUMER`, or drop the fallback and let unknowns hit `Default`.

---

src/PlanViewer.Core/Services/BenefitScorer.cs
```diff
@@ -38,10 +38,73 @@ public static void Score(ParsedPlan plan)
 
                 if (stmt.WaitStats.Count > 0 && stmt.QueryTimeStats != null)
                     ScoreWaitStats(stmt);
+
+                if (stmt.WaitStats.Count > 0)
+                    EmitWaitStatWarnings(stmt);
```
erikdarlingdata, line 43: `EmitWaitStatWarnings` runs even when `QueryTimeStats == null` (line 39's `ScoreWaitStats` is skipped in that case). With no `WaitBenefits` populated, `benefitByType` is empty and every wait gets `severity = Info` regardless of how dominant the wait is. That may be intentional (estimated/incomplete plans don't have elapsed to score against) but worth confirming — if it is, a one-line comment here saying "no QueryTimeStats → all severities pinned to Info" would save the next reader a trip through the lookup chain.

Also: the two adjacent `if (stmt.WaitStats.Count > 0)` blocks can collapse into a single guarded block, slightly cheaper and reads more clearly:
```csharp
if (stmt.WaitStats.Count > 0)
{
    if (stmt.QueryTimeStats != null) ScoreWaitStats(stmt);
    EmitWaitStatWarnings(stmt);
}
```

---

src/PlanViewer.Core/Services/BenefitScorer.cs
```diff
@@ -38,10 +38,73 @@ public static void Score(ParsedPlan plan)
...
+    private static void EmitWaitStatWarnings(PlanStatement stmt)
+    {
+        // Lookup benefit % by wait type (populated by ScoreWaitStats)
+        var benefitByType = new Dictionary<string, double>(StringComparer.OrdinalIgnoreCase);
+        foreach (var wb in stmt.WaitBenefits)
+            benefitByType[wb.WaitType] = wb.MaxBenefitPercent;
+
+        foreach (var wait in stmt.WaitStats)
+        {
+            if (wait.WaitTimeMs <= 0) continue;
```
erikdarlingdata, line 62: The only filter on emitting a warning is `WaitTimeMs <= 0`. Every nonzero wait — including 1ms `SOS_SCHEDULER_YIELD` chatter on a 30s query — becomes a first-class `PlanWarning` in the expander. On busy real-world plans (10–30 distinct wait types) this can flood the Plan Warnings UI with low-signal `Info` rows and make the genuinely actionable ones harder to find.

Consider a minimum-relevance threshold, e.g. skip emission if `wait.WaitTimeMs / elapsedMs < 0.005` and `benefitPct < 1`, or drop pure `Info` waits below some absolute floor.

---

erikdarlingdata (commented):
## Review summary

Release PR for v1.7.0 surfacing wait stats as first-class `PlanWarning`s. Four files, ~370 LOC, mostly the new `WaitStatsKnowledge` lookup table plus an `EmitWaitStatWarnings` pass in `BenefitScorer`. No XAML / Avalonia surface area changes, no MVVM creep, no contrast/brush regressions to flag.

### What this does
- `BenefitScorer.Score` gets a new pass that, for every `WaitStatInfo` on a statement, emits a `PlanWarning` with description + how-to-fix from `WaitStatsKnowledge`, severity bucketed by the per-wait `MaxBenefitPercent` from `ScoreWaitStats` (`>=50` Critical / `>=10` Warning / else Info).
- `WaitStatsKnowledge` is a static lookup with ~50 exact-match entries plus prefix fallbacks (LCK_M_, PAGEIOLATCH_, PAGELATCH_, LATCH_, HT, CX, PREEMPTIVE_) and a generic default. Never returns null.
- `PAGEIOLATCH_*` entries set `ShowEffectiveLatency = true`, so their warning message includes `wait_ms / wait_count` formatted via `FormatLatency`.
- Plumbed into the Web project via a `<Compile Include=... Link=...>` reference in `PlanViewer.Web.csproj` (correct pattern matching the rest of the file).

### What's good
- Lookup falls back gracefully and is null-safe.
- `FormatLatency` covers s/ms/µs ranges sensibly; division by `WaitCount` is guarded by `wait.WaitCount > 0`.
- Severity bucketing on `Nullable<double>` via pattern switch is clean — null benefit cleanly defaults to Info.
- Change is additive — no renderer/UI changes needed because everything flows through the existing `PlanWarnings` pipeline, which matches the PR description.

### What needs attention

1. **Base branch is `main`.** Project convention is to target `dev`. Confirmed this is intentional because it's a release PR (v1.6.0 → v1.7.0 in `PlanViewer.App.csproj`), but flagging per the standing review checklist.

2. **PerformanceMonitor sync.** This PR changes `BenefitScorer` (scoring) and adds the `WaitStatsKnowledge` rule data — both live in the analyzer surface that the **PerformanceMonitor Dashboard** and **PerformanceMonitor Lite** repos keep in sync with Studio. The same wait-warning emission + knowledge table needs to land there too, otherwise those products will diverge from Studio on what wait stats produce as warnings. (Also worth deciding whether the severity thresholds — `>=50` / `>=10` — should be a shared constant rather than duplicated in three repos.)

3. **No tests.** New logic in `src/PlanViewer.Core/Services/` (the entire `WaitStatsKnowledge.Lookup` fallback chain, severity bucketing, `FormatLatency`, message assembly) has no corresponding test in `tests/PlanViewer.Core.Tests/`. The prefix-fallback ordering and the `Exact["MEMORY_ALLOCATION_EXT"]` direct access (see inline) are exactly the kind of thing a 30-line table-driven test would have caught any regression on. At minimum: a few `[InlineData]` cases verifying exact match wins over prefix, `LATCH_FOO` vs `PAGELATCH_FOO` vs `PAGEIOLATCH_FOO` go to the right entry, unknown returns `Default`, severity buckets pick the right label.

4. **Inline issues** (left on the lines):
   - `WaitStatsKnowledge.cs:303` — `Exact["MEMORY_ALLOCATION_EXT"]` is a hard dictionary read in the fallback path; will throw if the key is ever removed.
   - `WaitStatsKnowledge.cs:281` and `:288` — `StartsWith("HT")` and `StartsWith("CX")` prefix fallbacks are too broad given the full known wait-type sets are already in `Exact`.
   - `BenefitScorer.cs:43` — `EmitWaitStatWarnings` runs even when `QueryTimeStats` is null, in which case all severities silently collapse to `Info`. Confirm intended.
   - `BenefitScorer.cs:62` — no minimum-benefit threshold; every nonzero wait emits a warning, which can flood the Plan Warnings expander on busy plans.

5. **Release-PR checklist** — the unchecked items in the description (web viewer + desktop expander visual checks post-deploy) are exactly the surfaces this change lights up. Worth doing before merging to main rather than after.

Not approving / not requesting changes — comments only.

---

PR #246, opened 2026-04-21
Make release workflow reruns idempotent when release already exists

erikdarlingdata:
## Summary
Only \`Create release\` retains the \`EXISTS == 'false'\` guard. Every downstream step (Setup .NET, Build, Publish, Sign, Velopack, Package/upload) always runs. \`gh release upload --clobber\` on the final step safely overwrites existing assets on rerun.

## Why
Happened during v1.7.0: first run timed out on SignPath approval after creating the release + tag, rerun skipped all work because v1.7.0 already existed. The only unblock was \`gh release delete v1.7.0 --cleanup-tag\` then rerun. After this change, a rerun after signing failure re-does the build + sign + upload and repopulates the empty release.

## Test plan
- [ ] Validate on the next dev→main release cycle — we're intentionally not reproducing the failure to test (rerunning this workflow on an already-shipped release would re-upload identical artifacts under \`--clobber\`, which is safe but noisy).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,20 +46,17 @@ jobs:
           gh release create "v${{ steps.version.outputs.VERSION }}" --title "v${{ steps.version.outputs.VERSION }}" --generate-notes --target main
 
       - name: Setup .NET 8.0
-        if: steps.check.outputs.EXISTS == 'false'
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: 8.0.x
 
       - name: Build and test
-        if: steps.check.outputs.EXISTS == 'false'
         run: |
           dotnet restore
           dotnet build -c Release
           dotnet test tests/PlanViewer.Core.Tests/PlanViewer.Core.Tests.csproj -c Release --no-build --verbosity normal
 
       - name: Publish App (all platforms)
-        if: steps.check.outputs.EXISTS == 'false'
         run: |
           dotnet publish src/PlanViewer.App/PlanViewer.App.csproj -c Release -r win-x64 --self-contained -o publish/win-x64
           dotnet publish src/PlanViewer.App/PlanViewer.App.csproj -c Release -r linux-x64 --self-contained -o publish/linux-x64
@@ -68,7 +65,6 @@ jobs:
 
       # ── SignPath code signing (Windows only, skipped if secret not configured) ──
       - name: Check if signing is configured
-        if: steps.check.outputs.EXISTS == 'false'
         id: signing
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -80,15 +76,15 @@ jobs:
           fi
 
       - name: Upload Windows build for signing
-        if: steps.check.outputs.EXISTS == 'false' && steps.signing.outputs.ENABLED == 'true'
+        if: steps.signing.outputs.ENABLED == 'true'
         id: upload-unsigned
         uses: actions/upload-artifact@v4
         with:
           name: App-unsigned
           path: publish/win-x64/
```
erikdarlingdata, line 84: `actions/upload-artifact@v4` requires unique artifact names within a workflow run. The target failure scenario here is "SignPath times out on attempt 1, user reruns" — if that rerun is done via GitHub's "Re-run failed jobs" (same run, attempt 2), the prior attempt's `App-unsigned` artifact persists and this step will fail with `an artifact with this name already exists`. Consider adding `overwrite: true` here to make the rerun path fully idempotent. A fresh workflow_dispatch run is unaffected, but "Re-run failed jobs" is the more likely maintainer action.

---

erikdarlingdata (commented):
**What it does**
Drops the `steps.check.outputs.EXISTS == 'false'` gate from every step after `Create release` (Setup .NET → Build/test → Publish → signing chain → Velopack → Package/upload). Only `Create release` itself keeps the guard. Reruns after the release+tag are created now re-execute build/sign/upload instead of no-opping.

**What's good**
- Base branch is `dev`. Correct.
- `gh release upload ... --clobber` at release.yml:184 makes the zip + SHA256SUMS upload safe to repeat.
- `vpk upload ... --merge` at release.yml:187 handles the Velopack side.
- Scope is minimal — no behavior change on the happy path.

**Needs attention**
- release.yml:78-84 — `actions/upload-artifact@v4` rejects duplicate artifact names within a run. If the user hits "Re-run failed jobs" (same run, attempt 2) after a SignPath timeout — which is exactly the scenario described in the PR body — the `App-unsigned` artifact from attempt 1 will collide. Adding `overwrite: true` closes this. A brand-new workflow run is unaffected, but the maintainer will likely reach for "Re-run failed jobs" first. Inline comment left.
- release.yml:117-120 — `vpk download` + `vpk pack` for the same `$VERSION` on a rerun: if a prior attempt already uploaded Velopack assets, `vpk download` will pull them back and treat v$VERSION as "previous" when packing v$VERSION. Not the failure mode this PR targets (sign-timeout halts before Velopack runs), so not blocking. Worth a quick sanity check that vpk tolerates that case before relying on rerun for failures past the Velopack step.
- Test plan is "validate on next release cycle" — fine, given the alternative (forcing a failure) would re-upload artifacts on a shipped release.

No C#/XAML changes, so Avalonia/brush/MVVM/test-coverage checks are N/A.

---

PR #247, opened 2026-04-21
Render benefit % and actionable fix in all warning surfaces (v1.7.1)

erikdarlingdata:
## Summary
Web viewer warnings strip was dropping \`MaxBenefitPercent\` and \`ActionableFix\` from every warning — the data flowed through \`ResultMapper\` into the JSON but the strip only rendered Severity/Type/Message. Joe caught the missing benefit % on v1.7.0 wait warnings.

Fixed the strip plus the same \`ActionableFix\` gap in the other surfaces that were already showing benefit (HTML export, Avalonia, plain text).

## Changes
- **Web viewer** (`Index.razor` + `app.css`): `.warn-benefit` badge + italic `.warning-fix` block. Strip now sorts by benefit desc, then severity, then type — matches HTML export and Avalonia.
- **HTML exporter**: appends `.warn-fix` block after `.warn-msg`.
- **Avalonia** (`PlanViewerControl`): italic TextBlock under the message when `ActionableFix` is present.
- **Text formatter**: indented \`Fix:\` line after each warning.
- **Version**: 1.7.0 → 1.7.1.

## Test plan
- [x] Text output verified — wait warnings now show \`(up to N% benefit)\` and \`Fix:\` line
- [ ] Web viewer visual check after deploy
- [ ] Avalonia Plan Warnings expander visual check after deploy

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #248, opened 2026-04-21
Release v1.7.1 — render benefit % and actionable fix

erikdarlingdata:
## Summary
Hotfix for the rendering gap Joe spotted on v1.7.0: the web warnings strip was dropping \`MaxBenefitPercent\` and \`ActionableFix\`. Fixed that, closed the same \`ActionableFix\` gap on HTML export / Avalonia / plain text, plus the idempotent-rerun workflow fix from PR #246.

See PR #247 and PR #246 for details.

## Test plan
- [x] Text output verified locally — benefit + Fix both present
- [ ] Web viewer visual check post-deploy
- [ ] Avalonia Plan Warnings expander visual check post-deploy

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,20 +46,17 @@ jobs:
           gh release create "v${{ steps.version.outputs.VERSION }}" --title "v${{ steps.version.outputs.VERSION }}" --generate-notes --target main
 
       - name: Setup .NET 8.0
-        if: steps.check.outputs.EXISTS == 'false'
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: 8.0.x
 
       - name: Build and test
-        if: steps.check.outputs.EXISTS == 'false'
         run: |
           dotnet restore
           dotnet build -c Release
           dotnet test tests/PlanViewer.Core.Tests/PlanViewer.Core.Tests.csproj -c Release --no-build --verbosity normal
 
       - name: Publish App (all platforms)
-        if: steps.check.outputs.EXISTS == 'false'
         run: |
           dotnet publish src/PlanViewer.App/PlanViewer.App.csproj -c Release -r win-x64 --self-contained -o publish/win-x64
           dotnet publish src/PlanViewer.App/PlanViewer.App.csproj -c Release -r linux-x64 --self-contained -o publish/linux-x64
@@ -68,7 +65,6 @@ jobs:
 
       # ── SignPath code signing (Windows only, skipped if secret not configured) ──
       - name: Check if signing is configured
-        if: steps.check.outputs.EXISTS == 'false'
         id: signing
         shell: bash
         run: |
@@ -80,7 +76,7 @@ jobs:
           fi
 
       - name: Upload Windows build for signing
-        if: steps.check.outputs.EXISTS == 'false' && steps.signing.outputs.ENABLED == 'true'
+        if: steps.signing.outputs.ENABLED == 'true'
         id: upload-unsigned
         uses: actions/upload-artifact@v4
         with:
@@ -88,7 +84,7 @@ jobs:
           path: publish/win-x64/
 
       - name: Sign Windows build
-        if: steps.check.outputs.EXISTS == 'false' && steps.signing.outputs.ENABLED == 'true'
+        if: steps.signing.outputs.ENABLED == 'true'
         uses: signpath/github-action-submit-signing-request@v1
         with:
           api-token: '${{ secrets.SIGNPATH_API_TOKEN }}'
@@ -101,7 +97,7 @@ jobs:
           output-artifact-directory: 'signed/win-x64'
 
       - name: Replace unsigned Windows build with signed
-        if: steps.check.outputs.EXISTS == 'false' && steps.signing.outputs.ENABLED == 'true'
+        if: steps.signing.outputs.ENABLED == 'true'
         shell: pwsh
         run: |
           Remove-Item -Recurse -Force publish/win-x64
@@ -109,7 +105,6 @@ jobs:
 
       # ── Velopack (uses signed Windows binaries) ───────────────────────
       - name: Create Velopack release (Windows)
-        if: steps.check.outputs.EXISTS == 'false'
         shell: pwsh
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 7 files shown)


src/PlanViewer.Core/Output/HtmlExporter.cs
```diff
@@ -459,6 +460,8 @@ private static void WriteWarnings(StringBuilder sb, StatementResult stmt)
             if (w.MaxBenefitPercent.HasValue)
                 sb.AppendLine($"<span class=\"warn-benefit\">up to {w.MaxBenefitPercent:N0}% benefit</span>");
             sb.AppendLine($"<span class=\"warn-msg\">{Encode(w.Message)}</span>");
+            if (!string.IsNullOrEmpty(w.ActionableFix))
+                sb.AppendLine($"<span class=\"warn-fix\">{Encode(w.ActionableFix)}</span>");
```
erikdarlingdata, line 464: New `warn-fix` span branch has no coverage in `tests/PlanViewer.Core.Tests/HtmlExporterTests.cs`. A snapshot-style assertion that a warning with `ActionableFix` set emits the span (and one with it null/empty does not) would guard against a regression like the one this PR is fixing.

---

src/PlanViewer.Core/Output/TextFormatter.cs
```diff
@@ -172,6 +172,8 @@ public static void WriteText(AnalysisResult result, TextWriter writer)
                         ? $" (up to {w.MaxBenefitPercent:N0}% benefit)"
                         : "";
                     writer.WriteLine($"  [{w.Severity}] {w.Type}{benefitTag}: {EscapeNewlines(w.Message)}");
+                    if (!string.IsNullOrEmpty(w.ActionableFix))
+                        writer.WriteLine($"    Fix: {EscapeNewlines(w.ActionableFix)}");
```
erikdarlingdata, line 176: Same gap on the text side — no test in `tests/PlanViewer.Core.Tests/` asserts the `    Fix: ...` line renders when `ActionableFix` is set. Text output is the easiest of the four surfaces to golden-test.

---

src/PlanViewer.Web/Pages/Index.razor
```diff
@@ -332,7 +332,10 @@ else
                 @if (infoCount > 0) { <span class="warn-count-badge info">@infoCount</span> }
             </h4>
             <div class="warnings-list">
-                @foreach (var w in GetAllWarnings(ActiveStmt!))
+                @foreach (var w in GetAllWarnings(ActiveStmt!)
+                    .OrderByDescending(x => x.MaxBenefitPercent ?? -1)
+                    .ThenByDescending(x => x.Severity == "Critical" ? 3 : x.Severity == "Warning" ? 2 : 1)
+                    .ThenBy(x => x.Type))
```
erikdarlingdata, line 338: Ordering here mirrors `TextFormatter.cs:166-168`, which is good — but the secondary key differs in direction. Text uses `.ThenBy(w => w.Severity switch { "Critical" => 0, "Warning" => 1, _ => 2 })` (ascending over an ordinal where Critical is lowest). Web uses `.ThenByDescending(x => x.Severity == "Critical" ? 3 : x.Severity == "Warning" ? 2 : 1)`. Both happen to yield Critical → Warning → Info, so behavior matches, but the inverted convention invites drift the next time someone touches one side and not the other. Worth normalizing to a single helper or identical expression.

---

.github/workflows/release.yml
```diff
@@ -126,7 +121,6 @@ jobs:
 
       # ── Package and upload ────────────────────────────────────────────
       - name: Package and upload
```
erikdarlingdata, line 123: Idempotency check: `gh release upload ... --clobber` (line 184) and `vpk upload ... --merge` (line 187) handle re-runs against an existing release, so dropping the `EXISTS == 'false'` guards on the build/sign/package steps is safe from a failure-mode standpoint. The trade-off is that every re-run now submits a fresh SignPath signing request and a full Velopack pack — intentional for the "partial failure on the first run" recovery case, just flagging the cost side for awareness.

---

erikdarlingdata (commented):
## What this does
v1.7.1 hotfix. `ActionableFix` was being dropped on the web warnings strip (Joe's report on v1.7.0); this PR renders it on web, HTML export, Avalonia properties panel, and plain-text output. Also backfills the `up to N% benefit` badge on the web strip, aligns web warning ordering with `TextFormatter`, and pulls in the idempotent-rerun workflow change from #246.

## What's good
- Rendering change is symmetric across all four output surfaces — text, HTML, web, Avalonia. Previous gap was web-only; this fix covers the set.
- Null-safe guards (`!string.IsNullOrEmpty(w.ActionableFix)`) on every surface — no empty `Fix:` rows when the field isn't populated.
- Avalonia addition at `PlanViewerControl.axaml.cs:1757-1766` uses the existing `TooltipFgBrush` (`#E4E6EB` on dark tooltip bg) — high contrast, no new dim-grey brush introduced.
- Web CSS uses existing `--text-secondary`/`--border` tokens rather than inline colors — no new contrast regressions.
- `gh release upload --clobber` and `vpk upload --merge` mean the workflow's dropped `EXISTS == 'false'` guards don't break re-runs.
- No PlanAnalyzer rule/scoring changes — no Monitor/Lite sync needed.

## What needs attention
- **No test coverage added.** `HtmlExporter.cs` and `TextFormatter.cs` both changed in `src/`, but `tests/PlanViewer.Core.Tests/HtmlExporterTests.cs` and the text-formatting tests weren't updated. Given the regression being fixed is exactly "output surface dropped a field," a golden-output test on each would be the cheapest insurance. See inline comments.
- **Sort expression drift between text and web** (`Index.razor:335-338` vs `TextFormatter.cs:166-168`). Same result today, opposite ordering direction on the severity key. Easy to diverge on the next change. Inline comment suggests consolidating.
- **Workflow cost side-effect**: removing the `EXISTS` guards means every workflow re-run now submits a full SignPath signing request and full Velopack pack, not just the upload. Intentional for the recovery case, worth being explicit about.

## Non-issues (pre-checked)
- `dev → main` base is correct — the release workflow at `release.yml:15` gates on `head.ref == 'dev'`, so this is the required pattern for a release PR.
- No AvaloniaEdit additions, no new `ComboBox`, no new `:pointerover` selectors, no MVVM creep.
- CSS vars used on web (`--text-secondary: #666666`) are darker than the rejected `#6B7280`/`#808080` range — contrast is fine.

---

PR #250, opened 2026-04-21
fix: bug-bash batch 2 — six small correctness/hygiene fixes

erikdarlingdata:
## Summary

Batch 2 from the bug-bash review — bundles six fixes that are all a few lines each, no feature changes.

| # | File | Fix |
|---|------|-----|
| 5 | `ReproScriptBuilder.cs` | Double `]` in DB names so `USE [cool]]stuff]` is well-formed instead of unbalanced. |
| 6 | `ReproScriptBuilder.cs` | Whitelist `SET TRANSACTION ISOLATION LEVEL` against the five valid T-SQL names; unknown/hostile strings drop silently. |
| 9 | `ShowPlanParser.cs` | `ParseLong` now passes `NumberStyles.Integer` + `InvariantCulture`, matching the sibling `ParseDouble`. |
| 10 | `KeychainCredentialService.cs` | Read stdout+stderr concurrently to avoid the classic sync-stderr-while-stdout-async deadlock (bites on `security dump-keychain` with large keychains). |
| 7 | `QueryStoreHistoryWindow.axaml.cs` | `OnClosed` override cancels + disposes `_fetchCts` so closing mid-load doesn't leave the SqlConnection open on the server. |
| 12 | `QuerySessionControl.axaml.cs` | `DetachedFromVisualTree` also cancels `_statusClearCts` so the fire-and-forget status-clear dispatch can't touch a detached control. |

## Test plan

- [x] `dotnet build PlanViewer.sln` — 0 errors.
- [x] **#5 (USE escape) — runnable repro:**
  - `databaseName = "cool]stuff"` → emits `USE [cool]]stuff];` ✅
  - `databaseName = "NormalDb"` → emits `USE [NormalDb];` (regression check) ✅
- [x] **#6 (isolation whitelist) — runnable repro:**
  - `"read committed"` → emits `SET TRANSACTION ISOLATION LEVEL READ COMMITTED;` ✅
  - `"READ COMMITTED; DROP DATABASE [prod];--"` → no SET line ✅
  - `"CHAOS"` → no SET line ✅
- [x] **#9 (ParseLong culture) — runnable repro:**
  - Under `de-DE`, invariant-digit string `"123456789012345"` parses correctly ✅
  - Overflow (`"99999999999999999999"`) still returns `0` — documented limitation, deliberately not changed in this PR to keep the diff tight.
- [x] **#7, #10, #12** — code review only (require live UI / macOS keychain). Each change is local and behaviorally narrow.

### DIFF
--- a/src/PlanViewer.App/Dialogs/QueryStoreHistoryWindow.axaml.cs
+++ b/src/PlanViewer.App/Dialogs/QueryStoreHistoryWindow.axaml.cs
@@ -169,6 +169,16 @@ public partial class QueryStoreHistoryWindow : Window
             : "AvgCpuMs";
     }
 
+    protected override void OnClosed(EventArgs e)
+    {
+        // Cancel any in-flight history fetch so the SqlConnection doesn't sit open on the
+        // server after the dialog is dismissed.
+        _fetchCts?.Cancel();
+        _fetchCts?.Dispose();
+        _fetchCts = null;
+        base.OnClosed(e);
+    }
+
     private async System.Threading.Tasks.Task LoadHistoryAsync()
     {
         _fetchCts?.Cancel();
--- a/src/PlanViewer.Core/Services/KeychainCredentialService.cs
+++ b/src/PlanViewer.Core/Services/KeychainCredentialService.cs
@@ -103,11 +103,14 @@ public class KeychainCredentialService : ICredentialService
         using var process = Process.Start(psi);
         if (process == null) return (-1, string.Empty);
 
+        // Read both streams concurrently. Doing stderr synchronously while stdout is
+        // async can deadlock if stderr fills its pipe buffer before the process exits
+        // (reproduces on `security dump-keychain` with large keychains).
         var stdoutTask = process.StandardOutput.ReadToEndAsync();
-        var stderr = process.StandardError.ReadToEnd();
+        var stderrTask = process.StandardError.ReadToEndAsync();
         process.WaitForExit();
-        var stdout = stdoutTask.Result;
+        Task.WaitAll(stdoutTask, stderrTask);
 
-        return (process.ExitCode, stdout + stderr);
+        return (process.ExitCode, stdoutTask.Result + stderrTask.Result);
     }
 }
--- a/src/PlanViewer.Core/Services/ReproScriptBuilder.cs
+++ b/src/PlanViewer.Core/Services/ReproScriptBuilder.cs
@@ -21,6 +21,20 @@ namespace PlanViewer.Core.Services;
 public static class ReproScriptBuilder
 {
     /// <summary>
+    /// Valid T-SQL isolation level names (uppercase) that may appear in a SET TRANSACTION
+    /// ISOLATION LEVEL statement. Used to gate interpolation so arbitrary upstream strings
+    /// can't land in the generated script.
+    /// </summary>
+    private static readonly HashSet<string> IsolationLevels = new(StringComparer.Ordinal)
+    {
+        "READ UNCOMMITTED",
+        "READ COMMITTED",
+        "REPEATABLE READ",
+        "SNAPSHOT",
+        "SERIALIZABLE",
+    };
+
+    /// <summary>
     /// Builds a complete reproduction script from available query data.
     /// </summary>
     public static string BuildReproScript(
@@ -94,10 +108,11 @@ public static class ReproScriptBuilder
         sb.AppendLine("*/");
         sb.AppendLine();
 
-        /* USE database (skip for Azure SQL DB — USE is invalid there) */
+        /* USE database (skip for Azure SQL DB — USE is invalid there).
+           Double any ']' in the identifier so names like 'cool]stuff' still parse. */
         if (!string.IsNullOrEmpty(databaseName) && !isAzureSqlDb)
         {
-            sb.AppendLine($"USE [{databaseName}];");
+            sb.AppendLine($"USE [{databaseName.Replace("]", "]]")}];");
             sb.AppendLine();
         }
 
@@ -116,7 +131,9 @@ public static class ReproScriptBuilder
 
         if (!string.IsNullOrEmpty(isolationLevel))
         {
-            sb.AppendLine($"SET TRANSACTION ISOLATION LEVEL {isolationLevel.ToUpperInvariant()};");
+            var upper = isolationLevel.ToUpperInvariant();
+            if (IsolationLevels.Contains(upper))
+                sb.AppendLine($"SET TRANSACTION ISOLATION LEVEL {upper};");
         }
         sb.AppendLine("SET NOCOUNT ON;");
         sb.AppendLine();
--- a/src/PlanViewer.Core/Services/ShowPlanParser.cs
+++ b/src/PlanViewer.Core/Services/ShowPlanParser.cs
@@ -1832,6 +1832,7 @@ public static class ShowPlanParser
     private static long ParseLong(string? value)
     {
         if (string.IsNullOrEmpty(value)) return 0;
-        return long.TryParse(value, out var result) ? result : 0;
+        return long.TryParse(value, System.Globalization.NumberStyles.Integer,
+            System.Globalization.CultureInfo.InvariantCulture, out var result) ? result : 0;
     }
 }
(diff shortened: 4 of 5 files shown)


src/PlanViewer.App/Dialogs/QueryStoreHistoryWindow.axaml.cs
```diff
@@ -169,6 +169,16 @@ public static string MapOrderByToMetricTag(string orderBy)
...
 
+    protected override void OnClosed(EventArgs e)
+    {
+        // Cancel any in-flight history fetch so the SqlConnection doesn't sit open on the
+        // server after the dialog is dismissed.
+        _fetchCts?.Cancel();
+        _fetchCts?.Dispose();
+        _fetchCts = null;
+        base.OnClosed(e);
+    }
```
erikdarlingdata, line 180: Pre-existing, not introduced here, but worth flagging while you're in this file: `LoadHistoryAsync` (lines 191–239) catches `OperationCanceledException` but the success path still touches `StatusText`, `HistoryDataGrid.ItemsSource`, `UpdateChart()`, and `PopulateLegendPanel()` without checking `ct.IsCancellationRequested`. If the fetch returns normally in the tiny window between `OnClosed` firing and the `await` resuming, those UI writes happen on a closed window. Consider a `ct.ThrowIfCancellationRequested()` after the `await` (or before the UI writes) so the post-await path routes through the `OperationCanceledException` catch.

---

src/PlanViewer.Core/Services/KeychainCredentialService.cs
```diff
@@ -103,11 +103,14 @@ private static (int ExitCode, string Output) RunSecurity(params string[] args)
...
 
+        // Read both streams concurrently. Doing stderr synchronously while stdout is
+        // async can deadlock if stderr fills its pipe buffer before the process exits
+        // (reproduces on `security dump-keychain` with large keychains).
         var stdoutTask = process.StandardOutput.ReadToEndAsync();
-        var stderr = process.StandardError.ReadToEnd();
+        var stderrTask = process.StandardError.ReadToEndAsync();
         process.WaitForExit();
-        var stdout = stdoutTask.Result;
+        Task.WaitAll(stdoutTask, stderrTask);
```
erikdarlingdata, line 113: Minor: `Task.WaitAll(stdoutTask, stderrTask)` after `process.WaitForExit()` is redundant — once the process has exited the pipes are closed and both `ReadToEndAsync` tasks will already be complete. Not a bug (it does guarantee exceptions surface rather than being lost on the task), just noting it. The Microsoft docs actually recommend the opposite order: start async reads, then `WaitForExit` (or `WaitForExitAsync`) — because `WaitForExit` without draining can itself block indefinitely in some shell scenarios. Current ordering works because both reads are async, but flipping to `await Task.WhenAll(stdoutTask, stderrTask); process.WaitForExit();` would be the textbook pattern.

---

src/PlanViewer.Core/Services/ReproScriptBuilder.cs
```diff
@@ -20,6 +20,20 @@ namespace PlanViewer.Core.Services;
...
+    /// ISOLATION LEVEL statement. Used to gate interpolation so arbitrary upstream strings
+    /// can't land in the generated script.
+    /// </summary>
+    private static readonly HashSet<string> IsolationLevels = new(StringComparer.Ordinal)
+    {
+        "READ UNCOMMITTED",
+        "READ COMMITTED",
+        "REPEATABLE READ",
+        "SNAPSHOT",
+        "SERIALIZABLE",
```
erikdarlingdata, line 34: Whitelist is correct per T-SQL spec — these are exactly the five values `SET TRANSACTION ISOLATION LEVEL` accepts. `READ COMMITTED SNAPSHOT` is deliberately absent and should be, since RCSI is a database-level option (`ALTER DATABASE … SET READ_COMMITTED_SNAPSHOT ON`) and not a transaction isolation level keyword. 👍

No tests in `tests/PlanViewer.Core.Tests/` cover `ReproScriptBuilder` whitelist or `]` escaping behavior — given this is now the security-hardening layer for a code-gen path, a couple of xunit cases pinning the two behaviors (valid level passes through, hostile string is silently dropped; `cool]stuff` → `cool]]stuff`) would be cheap insurance against regressions.

---

src/PlanViewer.Core/Services/ReproScriptBuilder.cs
```diff
@@ -94,10 +108,11 @@ public static string BuildReproScript(
         sb.AppendLine("*/");
         sb.AppendLine();
 
-        /* USE database (skip for Azure SQL DB — USE is invalid there) */
+        /* USE database (skip for Azure SQL DB — USE is invalid there).
+           Double any ']' in the identifier so names like 'cool]stuff' still parse. */
         if (!string.IsNullOrEmpty(databaseName) && !isAzureSqlDb)
         {
-            sb.AppendLine($"USE [{databaseName}];");
+            sb.AppendLine($"USE [{databaseName.Replace("]", "]]")}];");
```
erikdarlingdata, line 115: Escape is correct for the bracket-quoted identifier form. If anywhere else in this builder emits the database name without `[...]` brackets (e.g. as a plain token or inside a single-quoted string for `sp_executesql`), the `]→]]` doubling alone isn't the right escape for those forms. A quick grep for other `databaseName` usages in this file confirms whether this is the only emission site.

---

src/PlanViewer.Core/Services/ShowPlanParser.cs
```diff
@@ -1832,6 +1832,7 @@ private static double ParseDouble(string? value)
     private static long ParseLong(string? value)
     {
         if (string.IsNullOrEmpty(value)) return 0;
-        return long.TryParse(value, out var result) ? result : 0;
+        return long.TryParse(value, System.Globalization.NumberStyles.Integer,
+            System.Globalization.CultureInfo.InvariantCulture, out var result) ? result : 0;
```
erikdarlingdata, line 1836: Good alignment with `ParseDouble`. PR body notes the overflow case (`99999…`) still returns `0` — that's fine for ShowPlan's invariant-formatted payloads, just worth noting that silent-zero-on-overflow could mask a malformed plan. Not worth changing in this PR.

---

erikdarlingdata (commented):
## Review

**What it does:** Bug-bash batch 2 — five files, six narrow fixes: `]`-doubling for `USE [db]`, whitelist for `SET TRANSACTION ISOLATION LEVEL`, culture-invariant `ParseLong`, deadlock fix in `security` subprocess I/O, and cancel+dispose of two `CancellationTokenSource` fields on window close / control detach.

**Good:**
- Targets `dev` ✓
- Isolation-level whitelist uses the exact T-SQL-spec set with `StringComparer.Ordinal` + `ToUpperInvariant()` — correct.
- `]→]]` doubling is the correct escape for bracket-quoted identifiers.
- `ParseLong` now matches the `ParseDouble` sibling (`NumberStyles.Integer` + `InvariantCulture`) — no more `de-DE` surprises.
- Keychain deadlock fix is real; switching stderr from sync `ReadToEnd` to async solves the classic pipe-buffer deadlock.
- Both CTS-cleanup fixes (`QueryStoreHistoryWindow.OnClosed`, `QuerySessionControl.DetachedFromVisualTree`) follow the cancel-then-dispose-then-null pattern correctly.

**Needs attention (non-blocking, none regress existing behavior):**
- `QueryStoreHistoryWindow.LoadHistoryAsync` still has a post-await UI-write path that doesn't check `ct.IsCancellationRequested`. If the fetch returns between `OnClosed` and `await` resumption, `HistoryDataGrid.ItemsSource = …` runs on a closed window. Pre-existing; flagged since you're touching this lifecycle.
- `KeychainCredentialService.RunSecurity` ordering: Microsoft's documented pattern is `Task.WhenAll(reads); WaitForExit();` rather than `WaitForExit(); WaitAll(reads);`. Works either way with both streams async, but worth aligning.
- `ReproScriptBuilder` whitelist + `]`-escape are both new security-hardening surface area with no coverage in `tests/PlanViewer.Core.Tests/`. Two xunit cases would pin the hostile-string-drops and bracket-escape behaviors.

**Avalonia / theming:** No AvaloniaEdit, no `PlaceholderText`/`Watermark`, no `:pointerover` selectors, no new brushes or MVVM creep — nothing to flag.

**PlanAnalyzer:** Not touched. No Dashboard/Lite sync required.

Noted this was merged seven seconds after opening — comments are for the record / follow-up.

---

PR #255, opened 2026-04-22
Fix operator self-time through Compute Scalar children + one-decimal benefit % (#215 C5, C2)

erikdarlingdata:
## Summary
Two Joe-feedback items from #215:

- **C5 (bug):** operator self-time was wrong for any operator whose direct children were Compute Scalars. \`GetSerialOwnElapsed\` subtracted each child's \`ActualElapsedMs\`, but Compute Scalars don't carry runtime stats, so it subtracted 0 and the parent kept its full elapsed. Joe's node-6 Hash Spill reported 95% / 8,829ms; correct is 17.4% / 1,609ms (\`8.829 − 7.121 − 0.099\`). New \`GetEffectiveChildElapsedMs\` helper walks through pass-through operators to the first descendant with real stats.

- **C2:** benefit % was \`N0\` everywhere, so tiny waits showed as "up to 0%". Now one decimal except at 100 (shown as "100"). Applied to web strip + wait-stats card, HTML export, text formatter, Avalonia plan + node warning headers.

## Details
\`GetSerialOwnElapsed\` → iterates children calling \`GetEffectiveChildElapsedMs(child)\` instead of reading \`child.ActualElapsedMs\` directly. The helper:
- Parallelism exchange child → max-child elapsed (unchanged, exchange times unreliable)
- Child has \`ActualElapsedMs > 0\` → use it
- Otherwise → recurse into grandchildren and sum

Applies to every operator-time benefit scorer: spills (Rule 7), key lookups (Rule 10), scan with predicate (Rule 11), non-SARGable (Rule 12), bare scan (Rule 34), scan cardinality misestimate (Rule 32), eager index spool, filter operator, nested loops high executions, row estimate mismatch.

## Test plan
- [x] Joe's private plan — node-6 Hash Spill now shows 17.4% benefit, 1,609ms self-time (matches his math)
- [x] Wait items show 0.0–0.5% benefit instead of 0%
- [x] Regression check on \`20260415_1.sqlplan\` — all existing benefits unchanged
- [ ] Desktop Plan Warnings expander visual post-deploy

Version 1.7.1 → 1.7.2.

### DIFF
--- a/src/PlanViewer.App/Controls/PlanViewerControl.axaml.cs
+++ b/src/PlanViewer.App/Controls/PlanViewerControl.axaml.cs
@@ -801,6 +801,9 @@ public partial class PlanViewerControl : UserControl
         return $"{bytes / (1024L * 1024 * 1024):N1} GB";
     }
 
+    private static string FormatBenefitPercent(double pct) =>
+        pct >= 100 ? $"{pct:N0}" : $"{pct:N1}";
+
     #endregion
 
     #region Node Selection & Properties Panel
@@ -1737,7 +1740,7 @@ public partial class PlanViewerControl : UserControl
                         : w.Severity == PlanWarningSeverity.Warning ? "#FFB347" : "#6BB5FF";
                     var warnPanel = new StackPanel { Margin = new Thickness(10, 2, 10, 2) };
                     var planWarnHeader = w.MaxBenefitPercent.HasValue
-                        ? $"\u26A0 {w.WarningType} \u2014 up to {w.MaxBenefitPercent:N0}% benefit"
+                        ? $"\u26A0 {w.WarningType} \u2014 up to {FormatBenefitPercent(w.MaxBenefitPercent.Value)}% benefit"
                         : $"\u26A0 {w.WarningType}";
                     warnPanel.Children.Add(new TextBlock
                     {
@@ -1819,7 +1822,7 @@ public partial class PlanViewerControl : UserControl
                     : w.Severity == PlanWarningSeverity.Warning ? "#FFB347" : "#6BB5FF";
                 var warnPanel = new StackPanel { Margin = new Thickness(10, 2, 10, 2) };
                 var nodeWarnHeader = w.MaxBenefitPercent.HasValue
-                    ? $"\u26A0 {w.WarningType} \u2014 up to {w.MaxBenefitPercent:N0}% benefit"
+                    ? $"\u26A0 {w.WarningType} \u2014 up to {FormatBenefitPercent(w.MaxBenefitPercent.Value)}% benefit"
                     : $"\u26A0 {w.WarningType}";
                 warnPanel.Children.Add(new TextBlock
                 {
--- a/src/PlanViewer.Core/Output/HtmlExporter.cs
+++ b/src/PlanViewer.Core/Output/HtmlExporter.cs
@@ -407,7 +407,7 @@ pre.query-text, pre.text-output {
             {
                 var barPct = maxWait > 0 ? (double)w.WaitTimeMs / maxWait * 100 : 0;
                 var benefitTag = benefitLookup.TryGetValue(w.WaitType, out var pct)
-                    ? $" <span class=\"warn-benefit\">up to {pct:N0}%</span>"
+                    ? $" <span class=\"warn-benefit\">up to {(pct >= 100 ? pct.ToString("N0") : pct.ToString("N1"))}%</span>"
                     : "";
                 sb.AppendLine("<div class=\"wait-row\">");
                 sb.AppendLine($"<span class=\"wait-type\">{Encode(w.WaitType)}</span>");
@@ -458,7 +458,7 @@ pre.query-text, pre.text-output {
                 sb.AppendLine($"<span class=\"warn-op\">{Encode(w.Operator)}</span>");
             sb.AppendLine($"<span class=\"warn-type\">{Encode(w.Type)}</span>");
             if (w.MaxBenefitPercent.HasValue)
-                sb.AppendLine($"<span class=\"warn-benefit\">up to {w.MaxBenefitPercent:N0}% benefit</span>");
+                sb.AppendLine($"<span class=\"warn-benefit\">up to {(w.MaxBenefitPercent.Value >= 100 ? w.MaxBenefitPercent.Value.ToString("N0") : w.MaxBenefitPercent.Value.ToString("N1"))}% benefit</span>");
             sb.AppendLine($"<span class=\"warn-msg\">{Encode(w.Message)}</span>");
             if (!string.IsNullOrEmpty(w.ActionableFix))
                 sb.AppendLine($"<span class=\"warn-fix\">{Encode(w.ActionableFix)}</span>");
--- a/src/PlanViewer.Core/Services/PlanAnalyzer.cs
+++ b/src/PlanViewer.Core/Services/PlanAnalyzer.cs
@@ -1593,24 +1593,43 @@ public static class PlanAnalyzer
     }
 
     /// <summary>
-    /// Serial row mode self-time: subtract all direct children's elapsed.
-    /// Exchange children are skipped through to their real child.
+    /// Serial row mode self-time: subtract all direct children's effective elapsed.
+    /// Pass-through operators (Compute Scalar, etc.) don't carry runtime stats —
+    /// look through them to the first descendant that does. Exchange children
+    /// use max-child elapsed because exchange times are unreliable.
     /// </summary>
     private static long GetSerialOwnElapsed(PlanNode node)
     {
         var totalChildElapsed = 0L;
         foreach (var child in node.Children)
-        {
-            var childElapsed = child.ActualElapsedMs;
-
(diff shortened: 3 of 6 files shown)


src/PlanViewer.Core/Services/PlanAnalyzer.cs
```diff
@@ -1593,26 +1593,45 @@ private static long GetPerThreadOwnElapsed(PlanNode node)
...
+    /// <summary>
+    /// Returns the elapsed time a child contributes to its parent's subtree.
+    /// Looks through pass-through operators (Compute Scalar, Parallelism exchange)
+    /// that don't carry reliable runtime stats.
+    /// </summary>
+    private static long GetEffectiveChildElapsedMs(PlanNode child)
+    {
+        // Exchange operators: unreliable times, use max child
+        if (child.PhysicalOp == "Parallelism" && child.Children.Count > 0)
+            return child.Children.Max(GetEffectiveChildElapsedMs);
```
erikdarlingdata, line 1619: This is a scoring-rule change — `GetSerialOwnElapsed` feeds Rules 7, 10, 11, 12, 32, 34 plus eager index spool / filter / nested loops / row estimate scorers per the PR body. The equivalent fix needs to land in **PerformanceMonitor's Dashboard** and **Lite** copies of the analyzer or the three products will disagree on the same plan (Joe's node-6 Hash Spill would report 95% in Dashboard/Lite but 17.4% here).

---

src/PlanViewer.Core/Services/PlanAnalyzer.cs
```diff
@@ -1593,26 +1593,45 @@ private static long GetPerThreadOwnElapsed(PlanNode node)
...
+
+        // No stats (Compute Scalar and similar): look through to descendants
+        if (child.Children.Count == 0)
+            return 0;
+
+        var sum = 0L;
+        foreach (var grandchild in child.Children)
+            sum += GetEffectiveChildElapsedMs(grandchild);
+        return sum;
+    }
```
erikdarlingdata, line 1633: Summing over all grandchildren works for the common case (Compute Scalar with a single input) but could over-subtract if a pass-through-without-stats ever has multiple input branches — the parent's elapsed includes those branches sequentially in row mode, but the two sibling branches' elapsed are not additive in wall time. Compute Scalar is always unary so this is theoretical today; worth either a comment stating the assumption or a guard that only recurses when `child.Children.Count == 1`.

---

src/PlanViewer.Core/Services/PlanAnalyzer.cs
```diff
@@ -1593,26 +1593,45 @@ private static long GetPerThreadOwnElapsed(PlanNode node)
...
+    private static long GetEffectiveChildElapsedMs(PlanNode child)
+    {
+        // Exchange operators: unreliable times, use max child
+        if (child.PhysicalOp == "Parallelism" && child.Children.Count > 0)
+            return child.Children.Max(GetEffectiveChildElapsedMs);
+
+        // Child has its own stats: use them
+        if (child.ActualElapsedMs > 0)
+            return child.ActualElapsedMs;
+
```
erikdarlingdata, line 1624: Behavior change for Parallelism children worth calling out: the old code did `child.Children.Max(c => c.ActualElapsedMs)`, which would return 0 if the exchange's direct child was itself a Compute Scalar. The new recursive `Max(GetEffectiveChildElapsedMs)` now looks through that, which is probably what you want, but it changes the reported self-time for any operator sitting above a `Parallelism → Compute Scalar → Real Op` chain. Make sure the `20260415_1.sqlplan` regression check you ran exercises that shape.

---

src/PlanViewer.Core/Output/HtmlExporter.cs
```diff
@@ -407,7 +407,7 @@ private static void WriteWaitStatsCard(StringBuilder sb, StatementResult stmt, b
             {
                 var barPct = maxWait > 0 ? (double)w.WaitTimeMs / maxWait * 100 : 0;
                 var benefitTag = benefitLookup.TryGetValue(w.WaitType, out var pct)
-                    ? $" <span class=\"warn-benefit\">up to {pct:N0}%</span>"
+                    ? $" <span class=\"warn-benefit\">up to {(pct >= 100 ? pct.ToString("N0") : pct.ToString("N1"))}%</span>"
```
erikdarlingdata, line 410: The `>= 100 ? N0 : N1` ternary is now duplicated inline in HtmlExporter (x2), TextFormatter (x2), and Index.razor (x2), while PlanViewerControl has a `FormatBenefitPercent` helper for the same logic. Consider hoisting a single `BenefitFormatter.Format(double pct)` (or extension) into `PlanViewer.Core` so all six call sites agree — currently a future tweak (e.g. changing the cutoff to `>= 99.95`) has to be made in six places and could drift.

---

src/PlanViewer.App/Controls/PlanViewerControl.axaml.cs
```diff
@@ -801,6 +801,9 @@ private static string FormatBytes(double bytes)
         return $"{bytes / (1024L * 1024 * 1024):N1} GB";
     }
 
+    private static string FormatBenefitPercent(double pct) =>
+        pct >= 100 ? $"{pct:N0}" : $"{pct:N1}";
```
erikdarlingdata, line 805: Edge case: `pct` in the 99.95–99.99 range will format as `"100.0"` under `N1` (rounds up) but the ternary only switches to `N0` at `>= 100`. You'll get strings like `"up to 100.0%"` which reads weird next to the `"100"` case the branch is trying to produce. If the intent was "show integer only at exactly 100", consider rounding before the compare: `var r = Math.Round(pct, 1); return r >= 100 ? $"{r:N0}" : $"{r:N1}";`.

---

src/PlanViewer.Web/Pages/Index.razor
```diff
@@ -302,7 +302,7 @@ else
                                 @w.WaitTimeMs.ToString("N0") ms
                                 @if (benefitPct > 0)
                                 {
-                                    <span class="wait-benefit">up to @benefitPct.ToString("N0")%</span>
+                                    <span class="wait-benefit">up to @(benefitPct >= 100 ? benefitPct.ToString("N0") : benefitPct.ToString("N1"))%</span>
```
erikdarlingdata, line 305: Inline ternary inside the Razor expression is hard to read and duplicates the logic from `FormatBenefitPercent` in PlanViewerControl. Extracting to a `Core` helper (see HtmlExporter.cs comment) cleans this line up to `@BenefitFormatter.Format(benefitPct)%`.

---

erikdarlingdata (commented):
## Summary

Fixes two issues from #215 feedback:

1. **Self-time bug (C5):** `GetSerialOwnElapsed` subtracted raw `child.ActualElapsedMs`, which is 0 for Compute Scalars and similar pass-through operators, so the parent kept its full elapsed time. New `GetEffectiveChildElapsedMs` recurses through stats-less children to the first descendant with real timing data. Joe's node-6 Hash Spill goes from 95%/8,829ms to a correct 17.4%/1,609ms.
2. **Benefit % precision (C2):** `N0` formatting turned sub-1% benefits into "up to 0%". Now one decimal, reverting to integer at 100. Applied to Avalonia plan/node warning headers, Razor wait strip + warnings list, HtmlExporter, and TextFormatter.

Version bump 1.7.1 → 1.7.2. Targets `dev`.

## What's good

- Self-time fix is localized to one helper; serial-mode change is easy to reason about.
- Recursion is bounded by plan tree depth (no cycle risk).
- Parallelism handling is preserved and now also looks through chained pass-throughs.
- XML doc comments on both `GetSerialOwnElapsed` and `GetEffectiveChildElapsedMs` explain the intent.

## Needs attention

- **Cross-repo sync (PlanAnalyzer.cs:1605):** this is a behavior change to the scoring pipeline used by Rules 7/10/11/12/32/34 plus several operator scorers. **PerformanceMonitor Dashboard and Lite need the same fix** or the three products will disagree on the same plan.
- **Test coverage gap:** core analyzer logic changed but nothing in `tests/PlanViewer.Core.Tests/` was added/updated. A fixture for the "parent → Compute Scalar → Real Op" shape (and the Parallelism-over-Compute-Scalar shape) would lock in the fix and catch regressions. The PR body's regression check is manual against one `.sqlplan`.
- **Formatting duplication:** the `>= 100 ? N0 : N1` ternary is inlined in six places across three files while PlanViewerControl has a `FormatBenefitPercent` helper for the same thing. Worth hoisting a single formatter into `PlanViewer.Core`.
- **99.95 rounding edge (PlanViewerControl.axaml.cs:805):** `N1` rounds 99.95 up to "100.0", but the ternary only switches to `N0` at `>= 100`, so you can still get "up to 100.0%". Round before compare if the intent is "integer only at exactly 100".
- **Multi-child pass-through (PlanAnalyzer.cs:1633):** summing over all grandchildren is safe for unary Compute Scalars but would over-subtract for a hypothetical stats-less operator with multiple input branches in serial row mode. A `Children.Count == 1` guard or a comment stating the assumption would document the intent.

Inline notes left on the specific lines. Not approving or requesting changes — maintainer's call.

---

PR #256, opened 2026-04-22
Release v1.7.2 — operator self-time fix + decimal benefit %

erikdarlingdata:
## Summary
Hotfix for #215 Joe feedback:

- **C5**: operator self-time was wrong for parents of Compute Scalar children (bug since v1.5.0). Fixed via look-through helper. Node-6 Hash Spill on Joe's plan went from a bogus 95% to the correct 17.4%.
- **C2**: benefit % now shows one decimal instead of rounding tiny waits to "up to 0%".

See PR #255 for details.

## Test plan
- [x] Joe's private plan math matches
- [x] Regression check on \`20260415_1.sqlplan\`

### DIFF
--- a/server/PlanShare/Program.cs
+++ b/server/PlanShare/Program.cs
@@ -44,8 +44,14 @@ using (var conn = new SqliteConnection(connectionString))
     cmd.ExecuteNonQuery();
 }
 
+// --- Rate limiters (in-memory) ---
+// Created before Build() so they can be DI-registered and swept by CleanupService.
+var rateLimiter = new RateLimiter(maxRequests: 10, windowSeconds: 60);
+var analyticsRateLimiter = new RateLimiter(maxRequests: 30, windowSeconds: 60);
+
 // Register the cleanup background service
 builder.Services.AddSingleton(new PlanDbConfig(connectionString));
+builder.Services.AddSingleton(new RateLimiters(rateLimiter, analyticsRateLimiter));
 builder.Services.AddHostedService<CleanupService>();
 
 // Request size limit (10 MB)
@@ -53,10 +59,6 @@ builder.WebHost.ConfigureKestrel(o => o.Limits.MaxRequestBodySize = 10 * 1024 * 
 
 var app = builder.Build();
 app.UseCors();
-
-// --- Rate limiters (in-memory) ---
-var rateLimiter = new RateLimiter(maxRequests: 10, windowSeconds: 60);
-var analyticsRateLimiter = new RateLimiter(maxRequests: 30, windowSeconds: 60);
 
 const int MaxTtlDays = 365;
 
@@ -161,9 +163,15 @@ app.MapPost("/api/event", async (HttpContext ctx) =>
         return Results.BadRequest("Invalid JSON");
     }
 
-    // Strip referrer to domain only (no full URLs with query params)
-    if (!string.IsNullOrEmpty(referrer) && Uri.TryCreate(referrer, UriKind.Absolute, out var refUri))
-        referrer = refUri.Host;
+    // Strip referrer to domain only (no full URLs with query params).
+    // If it doesn't parse as an absolute URL, drop it — never persist raw
+    // client-supplied strings, since the dashboard renders referrers in HTML.
+    if (!string.IsNullOrEmpty(referrer))
+    {
+        referrer = Uri.TryCreate(referrer, UriKind.Absolute, out var refUri)
+            ? refUri.Host
+            : null;
+    }
 
     // Visitor hash: SHA256(IP + User-Agent + date) — unique per day, no PII stored
     var ua = ctx.Request.Headers.UserAgent.FirstOrDefault() ?? "";
@@ -352,14 +360,18 @@ static string GenerateDeleteToken()
 
 record PlanDbConfig(string ConnectionString);
 
+record RateLimiters(RateLimiter Share, RateLimiter Analytics);
+
 sealed class CleanupService : BackgroundService
 {
     private readonly PlanDbConfig _config;
+    private readonly RateLimiters _rateLimiters;
     private readonly ILogger<CleanupService> _logger;
 
-    public CleanupService(PlanDbConfig config, ILogger<CleanupService> logger)
+    public CleanupService(PlanDbConfig config, RateLimiters rateLimiters, ILogger<CleanupService> logger)
     {
         _config = config;
+        _rateLimiters = rateLimiters;
         _logger = logger;
     }
 
@@ -401,6 +413,14 @@ sealed class CleanupService : BackgroundService
                 if (deleted > 0)
                     _logger.LogInformation("Cleaned up {Count} old page views", deleted);
             }
+
+            // Evict stale rate-limiter keys so the dictionary doesn't grow forever.
+            var shareEvicted = _rateLimiters.Share.Sweep();
+            var analyticsEvicted = _rateLimiters.Analytics.Sweep();
+            if (shareEvicted + analyticsEvicted > 0)
+                _logger.LogInformation(
+                    "Evicted {Share} share + {Analytics} analytics rate-limit keys",
+                    shareEvicted, analyticsEvicted);
         }
         catch (Exception ex)
         {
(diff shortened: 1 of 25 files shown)


src/PlanViewer.Core/Services/PlanAnalyzer.cs
```diff
@@ -1593,26 +1593,45 @@ private static long GetPerThreadOwnElapsed(PlanNode node)
...
+        // No stats (Compute Scalar and similar): look through to descendants
+        if (child.Children.Count == 0)
+            return 0;
+
+        var sum = 0L;
+        foreach (var grandchild in child.Children)
+            sum += GetEffectiveChildElapsedMs(grandchild);
+        return sum;
+    }
+
```
erikdarlingdata, line 1634: Plan-analysis rule change. Confirm the equivalent look-through fix is landing in PerformanceMonitor's Dashboard and Lite copies — those repos must stay in sync with Studio's self-time math, otherwise the same Hash Spill / Compute Scalar parent will continue to report the stale ~95% figure there.

Minor: `GetEffectiveChildElapsedMs` treats `ActualElapsedMs == 0` as "no stats" and recurses, but a genuinely instantaneous operator with recorded 0ms would also recurse and over-count via grandchildren. Probably fine in practice (sub-ms timings don't aggregate into anything user-visible) but worth noting if you later hit a weird case.

---

server/PlanShare/Program.cs
```diff
@@ -441,4 +461,25 @@ public bool IsAllowed(string key)
...
+    public int Sweep()
+    {
+        var cutoff = DateTime.UtcNow.AddSeconds(-_windowSeconds);
+        var evicted = 0;
+        foreach (var kvp in _requests)
+        {
+            lock (kvp.Value)
+            {
+                kvp.Value.RemoveAll(t => t < cutoff);
+                if (kvp.Value.Count == 0 && _requests.TryRemove(kvp))
```
erikdarlingdata, line 479: Race with `IsAllowed`: a caller that already resolved `_requests.GetOrAdd(key, ...) → timestamps` can be blocked on `lock (timestamps)` while `Sweep` holds the same lock, observes `Count == 0`, and `TryRemove`s the kvp. When the caller then acquires the lock it adds its timestamp to a list that's no longer in the dictionary — the next `IsAllowed(key)` call creates a fresh list, so that one request effectively bypasses the rate limit.

Not catastrophic at 10/min, but worth either (a) re-checking that `_requests[key]` still points at `kvp.Value` after the remove, or (b) adding a sentinel/"tombstoned" flag on the list so a late arrival can detect the eviction and retry via `GetOrAdd`.

---

src/PlanViewer.Core/Services/WindowsCredentialService.cs
```diff
@@ -16,7 +16,7 @@ public bool SaveCredential(string serverId, string username, string password)
                 userName: username,
                 secret: password,
                 comment: "planview credential",
-                persistence: CredentialPersistence.LocalMachine);
+                persistence: CredentialPersistence.Enterprise);
```
erikdarlingdata, line 19: Persistence scope change from `LocalMachine` → `Enterprise` is a behavior change for existing installs. Credentials previously stored under the machine-wide scope won't be returned by a read that's targeting the enterprise (roaming) scope — upgrading users may silently lose saved SQL logins and be forced to re-enter them. If that's intentional (DPAPI scope hardening from #249), worth a release-notes line; if not, reading from both scopes on load and migrating once would avoid the regression.

---

src/PlanViewer.Cli/PasswordResolver.cs
```diff
@@ -0,0 +1,63 @@
...
+        if (passwordFromStdin)
+        {
+            if (!Console.IsInputRedirected)
+            {
+                Console.Error.WriteLine("--password-stdin requires stdin to be redirected (pipe the password into the command).");
+                return false;
+            }
+            password = Console.In.ReadLine()?.TrimEnd('\r', '\n') ?? "";
+            return true;
+        }
```
erikdarlingdata, line 49: `Console.In.ReadLine()` already strips the trailing line terminator, so the `.TrimEnd('\r', '\n')` is redundant — and if a caller pipes a password that legitimately ends in `\r` (unlikely but possible on Windows-CRLF input that's been partially normalized), you'd silently strip a real character. Safe to drop the TrimEnd.

---

src/PlanViewer.App/Services/AtomicFile.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// Writes <paramref name="contents"/> to <paramref name="path"/> atomically
+    /// with respect to process crashes. If the process dies before the rename,
+    /// <paramref name="path"/> keeps its previous contents and a stray
+    /// <c>.tmp</c> sibling is left behind (cleaned up on the next call).
+    /// </summary>
+    public static void WriteAllText(string path, string contents)
+    {
+        var tmp = path + ".tmp";
+        File.WriteAllText(tmp, contents);
+        // File.Move with overwrite:true maps to MoveFileEx(MOVEFILE_REPLACE_EXISTING)
```
erikdarlingdata, line 22: Two concurrent writers (two Studio processes saving settings at once) both race on `path + ".tmp"` — `File.WriteAllText` on the second one truncates the first's in-progress temp, then both `File.Move` calls succeed in some order, so the final content is well-formed but may be either process's payload rather than the later write. Settings don't change often enough for this to matter, but the docstring's "atomic with respect to process crashes" is accurate — not "atomic with respect to concurrent writers." A unique suffix (e.g. `path + "." + Guid.NewGuid().ToString("N") + ".tmp"`) would close it if you ever need it.

---

erikdarlingdata (commented):
### What this PR does
Release v1.7.2 rollup. Two user-visible fixes from Joe's feedback on #215:
- **C5 operator self-time**: `PlanAnalyzer.GetSerialOwnElapsed` now looks through pass-through operators (Compute Scalar, Parallelism exchange) that don't carry reliable runtime stats, via a new `GetEffectiveChildElapsedMs` helper. The Hash Spill-parent-of-Compute-Scalar case that was reporting ~95% self-time now reports the correct ~17%.
- **C2 benefit %**: wait-stat / warning "up to X%" headers render with one decimal below 100 via a `FormatBenefitPercent` helper, so tiny waits no longer round to "up to 0%." Applied consistently across `PlanViewerControl`, `HtmlExporter`, `TextFormatter`, and `Web/Index.razor`.

Also bundles previously-merged fixes: dashboard XSS hardening, rate-limiter key eviction, atomic settings writes, random SSMS temp-plan filename, `--password-stdin`, keychain stderr-deadlock fix, isolation-level allowlist in `ReproScriptBuilder`, culture-invariant `ParseLong`, `CredentialPersistence.Enterprise`.

### Good
- Benefit-% helper is applied everywhere the old `N0` format appeared — no surface missed.
- Sweep() is wired through DI and invoked by `CleanupService` rather than a separate timer; nice.
- Dashboard switches from `innerHTML` to `textContent` + `createElement` — correct XSS fix.
- `ReproScriptBuilder` isolation-level allowlist and `databaseName` `]`-doubling close two real SQL-injection shapes in a user-facing script generator.
- `KeychainCredentialService` concurrent stream read fixes a real deadlock.

### Needs attention
1. **Base branch is `main`**. Typical release flow, but flagging per convention — non-release PRs should target `dev`.
2. **PlanAnalyzer rule change must mirror into PerformanceMonitor Dashboard + Lite.** Those separate repos share Studio's rule set; if they don't get the same `GetEffectiveChildElapsedMs` update they'll keep reporting the bogus self-time. Inline on `PlanAnalyzer.cs:1634`.
3. **Rate-limiter Sweep race.** `IsAllowed` can hold a reference to a list that `Sweep` just evicted from the dictionary; the next timestamp lands on an orphaned list. Inline on `Program.cs:479`.
4. **Credential persistence change is breaking for existing users.** `LocalMachine` → `Enterprise` won't read back credentials written under the prior scope. Inline on `WindowsCredentialService.cs:19`.
5. **No test changes in `tests/PlanViewer.Core.Tests/`.** The self-time fix is exactly the kind of thing that should land with a regression test — e.g. a `.sqlplan` fixture reproducing the Hash-Spill-over-Compute-Scalar case, asserting `GetSerialOwnElapsed` returns the corrected value. `PasswordResolver`, `AtomicFile`, `ReproScriptBuilder` isolation gating, and `ParseLong` invariant-culture all lack coverage too.
6. Nit on `PasswordResolver` line 49 — redundant `TrimEnd` after `ReadLine`.
7. Nit on `AtomicFile` — crash-safe but not safe against two concurrent writers to the same path.

---

PR #257, opened 2026-04-22
Show deployed version on the web viewer

erikdarlingdata:
## Summary
Renders the deployed PlanViewer.Web assembly version on two surfaces so Joe (and anyone reviewing) can tell at a glance which release is live:
- Landing page: small badge on the "Powered by Performance Studio" line
- Toolbar: badge on the right when a plan is loaded

## Why the small pile of changes
- \`Index.razor\`: adds \`AppVersion\` helper reading \`typeof(Index).Assembly.GetName().Version\`
- \`app.css\`: adds \`.toolbar-app-version\` and \`.app-version\` classes
- \`PlanViewer.Web.csproj\`: one-time bump 1.4.0 → 1.7.3 so the assembly version matches the App release
- \`deploy-web.yml\`: before publish, reads App.csproj \`Version\` and writes it into Web.csproj so we never have to remember to bump both. Also broadened the path trigger to include App.csproj so pure version-bump releases still redeploy the web.
- \`PlanViewer.App.csproj\`: version bump 1.7.2 → 1.7.3.

## Test plan
- [x] Debug build of web project green
- [ ] Visual check post-deploy: badge readable on both landing and toolbar

### DIFF
--- a/.github/workflows/deploy-web.yml
+++ b/.github/workflows/deploy-web.yml
@@ -6,6 +6,7 @@ on:
     paths:
       - 'src/PlanViewer.Core/**'
       - 'src/PlanViewer.Web/**'
+      - 'src/PlanViewer.App/PlanViewer.App.csproj'
       - '.github/workflows/deploy-web.yml'
   workflow_dispatch:
 
@@ -35,6 +36,17 @@ jobs:
 
       - name: Install WASM workload
         run: dotnet workload install wasm-tools
+
+      - name: Sync web version with app version
+        shell: pwsh
+        run: |
+          $appVersion = ([xml](Get-Content src/PlanViewer.App/PlanViewer.App.csproj)).Project.PropertyGroup.Version | Where-Object { $_ }
+          Write-Host "App version: $appVersion"
+          $webCsproj = 'src/PlanViewer.Web/PlanViewer.Web.csproj'
+          $content = Get-Content $webCsproj -Raw
+          $content = [regex]::Replace($content, '<Version>[^<]+</Version>', "<Version>$appVersion</Version>")
+          Set-Content -Path $webCsproj -Value $content -NoNewline
+          Write-Host "Synced web csproj to $appVersion"
 
       - name: Publish Blazor WASM
         run: dotnet publish src/PlanViewer.Web/PlanViewer.Web.csproj -c Release -o publish
--- a/src/PlanViewer.Web/Pages/Index.razor
+++ b/src/PlanViewer.Web/Pages/Index.razor
@@ -7,7 +7,7 @@
         <div class="landing-content">
             <h2>Free SQL Server Query Plan Analysis</h2>
             <p class="privacy">Paste or upload a .sqlplan file. Your plan XML never leaves your browser.</p>
-            <p class="powered-by">Powered by the same analysis engine in the <a href="https://github.com/erikdarlingdata/PerformanceStudio" target="_blank" rel="noopener">Performance Studio</a> app.</p>
+            <p class="powered-by">Powered by the same analysis engine in the <a href="https://github.com/erikdarlingdata/PerformanceStudio" target="_blank" rel="noopener">Performance Studio</a> app. <span class="app-version">@AppVersion</span></p>
 
             @if (errorMessage != null)
             {
@@ -57,6 +57,7 @@ else
         {
             <span class="toolbar-version">@result.SqlServerBuild</span>
         }
+        <span class="toolbar-app-version" title="Performance Studio web viewer version">@AppVersion</span>
         @if (shareUrl != null)
         {
             <span class="share-url"><a href="@shareUrl" target="_blank">@shareUrl</a></span>
@@ -1663,6 +1664,13 @@ else
 
 @code {
     private const string ShareApiBase = "https://stats.erikdarling.com";
+
+    private static readonly string AppVersion = GetAppVersion();
+    private static string GetAppVersion()
+    {
+        var v = typeof(Index).Assembly.GetName().Version;
+        return v == null ? "?" : $"v{v.Major}.{v.Minor}.{v.Build}";
+    }
 
     private string activeTab = "paste";
     private string planXml = "";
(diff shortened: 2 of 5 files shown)


.github/workflows/deploy-web.yml
```diff
@@ -36,6 +37,17 @@ jobs:
       - name: Install WASM workload
         run: dotnet workload install wasm-tools
 
+      - name: Sync web version with app version
+        shell: pwsh
+        run: |
+          $appVersion = ([xml](Get-Content src/PlanViewer.App/PlanViewer.App.csproj)).Project.PropertyGroup.Version | Where-Object { $_ }
```
erikdarlingdata, line 43: `([xml]...).Project.PropertyGroup.Version | Where-Object { $_ }` assumes `PlanViewer.App.csproj` has exactly one `<PropertyGroup>` carrying `<Version>`. If a conditional second PropertyGroup is ever added (common when adding `<PropertyGroup Condition="...">` blocks), `.PropertyGroup` becomes an array, `.Version` becomes a collection, and `$appVersion` interpolates as a space-separated string into the regex replace — producing an invalid `<Version>1.7.3 </Version>` or similar. Safer: `Select-Xml -Path ... -XPath '/Project/PropertyGroup/Version[1]/text()' | ForEach-Object { $_.Node.Value }`.

---

.github/workflows/deploy-web.yml
```diff
@@ -36,6 +37,17 @@ jobs:
...
+          Write-Host "App version: $appVersion"
+          $webCsproj = 'src/PlanViewer.Web/PlanViewer.Web.csproj'
+          $content = Get-Content $webCsproj -Raw
+          $content = [regex]::Replace($content, '<Version>[^<]+</Version>', "<Version>$appVersion</Version>")
+          Set-Content -Path $webCsproj -Value $content -NoNewline
```
erikdarlingdata, line 48: `Set-Content -NoNewline` strips the trailing newline the csproj currently has. Harmless in CI since the file isn't committed back, but if anyone ever commits the rewritten file (e.g. a future step that does `git add`), they'll see a spurious EOL-only diff. Drop `-NoNewline` or append `[Environment]::NewLine` to `$content`.

---

src/PlanViewer.Web/wwwroot/css/app.css
```diff
@@ -451,6 +451,23 @@ textarea::placeholder {
...
 }
 
+.toolbar-app-version {
+    font-size: 0.75rem;
+    font-weight: 600;
+    color: var(--text-muted);
+    padding: 0.1rem 0.4rem;
+    border-radius: 3px;
+    background: rgba(0, 0, 0, 0.05);
+    margin-left: auto;
```
erikdarlingdata, line 461: `margin-left: auto` on `.toolbar-app-version` pushes the badge right, but `.toolbar` is `display: flex` with `.share-url` and `.delete-share-btn` rendered **after** the badge in the markup (`Index.razor:60-71`). Those siblings get carried along to the right edge too, so the badge ends up left-of the share URL rather than solo on the right. If the intent (per PR body: "badge on the right when a plan is loaded") is for the badge to be the rightmost element, move the `<span class="toolbar-app-version">` after the share-url/delete-btn blocks in the razor, or put `margin-left: auto` on whichever element should be the left edge of the right-aligned group.

---

src/PlanViewer.Web/Pages/Index.razor
```diff
@@ -1664,6 +1665,13 @@ else
 @code {
     private const string ShareApiBase = "https://stats.erikdarling.com";
 
+    private static readonly string AppVersion = GetAppVersion();
+    private static string GetAppVersion()
+    {
+        var v = typeof(Index).Assembly.GetName().Version;
+        return v == null ? "?" : $"v{v.Major}.{v.Minor}.{v.Build}";
```
erikdarlingdata, line 1672: `v.Build` is the third component of `System.Version` (Major.Minor.Build.Revision), which for a `<Version>1.7.3</Version>` csproj resolves to AssemblyVersion `1.7.3.0` and renders `v1.7.3` — correct. Worth noting: if the csproj ever uses a 4-part `<Version>` (e.g. `1.7.3.4`), the Revision is silently dropped. Probably fine given the sync script writes 3-part versions, but flag if that assumption ever loosens.

---

erikdarlingdata (commented):
**What it does**
- Surfaces the deployed `PlanViewer.Web` assembly version as a badge on the landing page ("Powered by..." line) and on the toolbar when a plan is loaded.
- Adds a CI step in `deploy-web.yml` that rewrites `<Version>` in `PlanViewer.Web.csproj` from `PlanViewer.App.csproj` before publish, so the two can't drift.
- Bumps `PlanViewer.App` 1.7.2 → 1.7.3 and one-time-syncs `PlanViewer.Web` 1.4.0 → 1.7.3 in source.
- Broadens the workflow `paths:` trigger to include `PlanViewer.App.csproj` so pure version-bump releases redeploy the web.

**What's good**
- Targets `dev`, not `main`.
- Version-drift problem solved at the CI boundary rather than by process/discipline.
- Both new CSS classes reuse `var(--text-muted)` instead of hardcoding a dim hex — good.
- `AppVersion` helper handles a null `.Version` with a `"?"` fallback.

**What needs attention**
- PowerShell XML read in `deploy-web.yml` is brittle if `PlanViewer.App.csproj` ever gains a second `<PropertyGroup>` — see inline comment.
- `Set-Content -NoNewline` silently strips the trailing newline on the Web csproj — inline comment.
- `margin-left: auto` on `.toolbar-app-version` will carry share-url/delete-btn to the right with it, so the badge isn't solo-rightmost when those are rendered. Confirm intent — inline comment on `app.css:461`.
- Minor: `v.Build` assumption on a 3-part `<Version>` works today but silently drops a 4th component if ever used.

**Not applicable / clean**
- No Avalonia surface touched (no AvaloniaEdit/ComboBox/`:pointerover`/code-behind concerns).
- No PlanAnalyzer rule or scoring changes — no Dashboard/Lite sync needed.
- No logic in `src/` beyond a display helper, so no missing test coverage to flag.
- No swallowed exceptions, off-thread UI work, or post-dispose usage.

---

PR #258, opened 2026-04-22
Release v1.7.3 — visible web viewer version

erikdarlingdata:
## Summary
- Web viewer now shows its deployed version on the landing page and toolbar so reviewers (Joe) can tell at a glance which code they're testing against.
- \`deploy-web.yml\` auto-syncs the web csproj version with the app csproj version and triggers on App.csproj changes so releases keep the web deploy in sync.

See PR #257.

## Test plan
- [x] PR #257 green
- [ ] Visual post-deploy: badge visible on landing + toolbar, version reads v1.7.3

### DIFF
--- a/.github/workflows/deploy-web.yml
+++ b/.github/workflows/deploy-web.yml
@@ -6,6 +6,7 @@ on:
     paths:
       - 'src/PlanViewer.Core/**'
       - 'src/PlanViewer.Web/**'
+      - 'src/PlanViewer.App/PlanViewer.App.csproj'
       - '.github/workflows/deploy-web.yml'
   workflow_dispatch:
 
@@ -35,6 +36,17 @@ jobs:
 
       - name: Install WASM workload
         run: dotnet workload install wasm-tools
+
+      - name: Sync web version with app version
+        shell: pwsh
+        run: |
+          $appVersion = ([xml](Get-Content src/PlanViewer.App/PlanViewer.App.csproj)).Project.PropertyGroup.Version | Where-Object { $_ }
+          Write-Host "App version: $appVersion"
+          $webCsproj = 'src/PlanViewer.Web/PlanViewer.Web.csproj'
+          $content = Get-Content $webCsproj -Raw
+          $content = [regex]::Replace($content, '<Version>[^<]+</Version>', "<Version>$appVersion</Version>")
+          Set-Content -Path $webCsproj -Value $content -NoNewline
+          Write-Host "Synced web csproj to $appVersion"
 
       - name: Publish Blazor WASM
         run: dotnet publish src/PlanViewer.Web/PlanViewer.Web.csproj -c Release -o publish
--- a/src/PlanViewer.Web/Pages/Index.razor
+++ b/src/PlanViewer.Web/Pages/Index.razor
@@ -7,7 +7,7 @@
         <div class="landing-content">
             <h2>Free SQL Server Query Plan Analysis</h2>
             <p class="privacy">Paste or upload a .sqlplan file. Your plan XML never leaves your browser.</p>
-            <p class="powered-by">Powered by the same analysis engine in the <a href="https://github.com/erikdarlingdata/PerformanceStudio" target="_blank" rel="noopener">Performance Studio</a> app.</p>
+            <p class="powered-by">Powered by the same analysis engine in the <a href="https://github.com/erikdarlingdata/PerformanceStudio" target="_blank" rel="noopener">Performance Studio</a> app. <span class="app-version">@AppVersion</span></p>
 
             @if (errorMessage != null)
             {
@@ -57,6 +57,7 @@ else
         {
             <span class="toolbar-version">@result.SqlServerBuild</span>
         }
+        <span class="toolbar-app-version" title="Performance Studio web viewer version">@AppVersion</span>
         @if (shareUrl != null)
         {
             <span class="share-url"><a href="@shareUrl" target="_blank">@shareUrl</a></span>
@@ -1663,6 +1664,13 @@ else
 
 @code {
     private const string ShareApiBase = "https://stats.erikdarling.com";
+
+    private static readonly string AppVersion = GetAppVersion();
+    private static string GetAppVersion()
+    {
+        var v = typeof(Index).Assembly.GetName().Version;
+        return v == null ? "?" : $"v{v.Major}.{v.Minor}.{v.Build}";
+    }
 
     private string activeTab = "paste";
     private string planXml = "";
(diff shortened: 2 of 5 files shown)


.github/workflows/deploy-web.yml
```diff
@@ -36,6 +37,17 @@ jobs:
       - name: Install WASM workload
         run: dotnet workload install wasm-tools
 
+      - name: Sync web version with app version
+        shell: pwsh
+        run: |
+          $appVersion = ([xml](Get-Content src/PlanViewer.App/PlanViewer.App.csproj)).Project.PropertyGroup.Version | Where-Object { $_ }
+          Write-Host "App version: $appVersion"
+          $webCsproj = 'src/PlanViewer.Web/PlanViewer.Web.csproj'
+          $content = Get-Content $webCsproj -Raw
```
erikdarlingdata, line 46: `.Project.PropertyGroup.Version` is fragile: when a csproj has multiple `<PropertyGroup>` blocks (e.g. a Debug/Release conditional group), `PropertyGroup` becomes an array and `.Version` returns an array of values. `Where-Object { $_ }` filters empty strings but not duplicates — if a future PropertyGroup also declares `<Version>`, `$appVersion` becomes an array and the regex replacement below interpolates something like `System.Object[]` into the web csproj. Consider `Select-Xml -XPath '/Project/PropertyGroup/Version' | Select-Object -First 1 -ExpandProperty Node | ForEach-Object InnerText` or similar, and assert a single scalar result before proceeding.

---

.github/workflows/deploy-web.yml
```diff
@@ -36,6 +37,17 @@ jobs:
...
+          $content = [regex]::Replace($content, '<Version>[^<]+</Version>', "<Version>$appVersion</Version>")
+          Set-Content -Path $webCsproj -Value $content -NoNewline
+          Write-Host "Synced web csproj to $appVersion"
```
erikdarlingdata, line 49: The regex `<Version>[^<]+</Version>` is unanchored, so if a `<PackageReference ... Version="...">` ever gets rewritten to the child-element form (`<PackageReference ...><Version>...</Version></PackageReference>`), or any other `<Version>` element is added, this replaces all of them. Safer to anchor inside a PropertyGroup or load the csproj as XML and set the node directly, matching how `$appVersion` is read.

---

src/PlanViewer.Web/PlanViewer.Web.csproj
```diff
@@ -5,7 +5,7 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <RootNamespace>PlanViewer.Web</RootNamespace>
-    <Version>1.4.0</Version>
+    <Version>1.7.3</Version>
```
erikdarlingdata, line 8: Now that the workflow rewrites this value at publish time, the committed value is purely decorative. Worth a one-line comment here (or in the sync step) noting that the source of truth is `PlanViewer.App.csproj` so the next person bumping the app version doesn't also update this by hand and get confused when CI still overwrites it.

---

src/PlanViewer.Web/Pages/Index.razor
```diff
@@ -1664,6 +1665,13 @@ else
 @code {
     private const string ShareApiBase = "https://stats.erikdarling.com";
 
+    private static readonly string AppVersion = GetAppVersion();
+    private static string GetAppVersion()
+    {
+        var v = typeof(Index).Assembly.GetName().Version;
```
erikdarlingdata, line 1671: `Assembly.GetName().Version` returns the AssemblyVersion, which defaults to `Major.Minor.Build.0` when only `<Version>` is set — so `1.7.3` → `1.7.3.0` and `Major.Minor.Build` renders `1.7.3` correctly. Just note that if anyone later sets `<AssemblyVersion>` explicitly (e.g. `1.0.0.0` for binary-compat reasons, common in libraries), this falls out of sync with the displayed version. `AssemblyInformationalVersionAttribute` (read via `Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion`) tracks `<Version>` directly and is the more robust source for user-facing version strings.

---

src/PlanViewer.Web/wwwroot/css/app.css
```diff
@@ -451,6 +451,23 @@ textarea::placeholder {
     color: var(--text-muted);
 }
 
+.toolbar-app-version {
+    font-size: 0.75rem;
+    font-weight: 600;
+    color: var(--text-muted);
+    padding: 0.1rem 0.4rem;
+    border-radius: 3px;
+    background: rgba(0, 0, 0, 0.05);
```
erikdarlingdata, line 460: Both badges bind `color: var(--text-muted)`, which resolves to `#999999` (app.css:12). That's in the dim-grey range the desktop side explicitly rejected (ForegroundMutedBrush was moved to `#B0B6C0` for readability). `.toolbar-app-version` additionally puts that grey on a `rgba(0,0,0,0.05)` tint, which reduces contrast further on light toolbars. Not introduced by this PR — the variable is already widespread — but the new badges are intentionally a "reviewers need to see the version at a glance" affordance, so it's worth either bumping them up (e.g. closer to `--text-secondary`/primary) or giving the pill a stronger background. Otherwise the thing that's supposed to be visible is the hardest-to-read label on the page.

---

erikdarlingdata (commented):
## Summary

v1.7.3 release PR: bumps `PlanViewer.App` version, adds a visible version badge to the web viewer's landing page and toolbar, and teaches `deploy-web.yml` to auto-sync the web csproj version from the app csproj (plus adds the app csproj to the web deploy's path triggers).

## Base branch

**Base is `main`, not `dev`.** Flagging per repo convention — this is clearly a release merge (dev → main) so it's probably intentional, but worth confirming you meant to merge `dev` forward rather than land this on `dev` first.

## What's good

- Path trigger for `PlanViewer.App.csproj` fixes the actual problem: version-only bumps were not redeploying the web viewer, so the displayed version would have lagged.
- Single source of truth (App csproj) with a CI-side sync is the right shape — avoids the double-bump drift the web csproj was already showing (`1.4.0` vs `1.7.2`).
- `AppVersion` computed in a `static readonly` with a private helper — cheap, no per-render reflection.
- No Avalonia / PlanAnalyzer / `src/` logic changes, so no Dashboard/Lite sync or test-coverage concerns apply.

## Needs attention

- **CI sync step is fragile**: XML access via `.Project.PropertyGroup.Version` and the unanchored regex replacement both break in plausible future csproj shapes. Inline comments on `deploy-web.yml` lines 46 and 49.
- **Version display via `AssemblyName.Version`** silently falls out of sync if anyone later sets `<AssemblyVersion>` explicitly. `AssemblyInformationalVersion` is more robust. Inline comment on `Index.razor:1671`.
- **Badge contrast**: both new CSS classes use `--text-muted` (`#999999`), and `.toolbar-app-version` layers it on a faint dark tint. Given the badge exists specifically to be readable at a glance, consider a higher-contrast color or a stronger pill background. Inline comment on `app.css:454`.
- **Stale committed value in `PlanViewer.Web.csproj`**: now that CI overwrites it, flag it in-tree so a future bumper doesn't chase it by hand.

Comments only — not approving, not requesting changes.

---

PR #259, opened 2026-04-22
External/preemptive wait benefit formula + CPU:Elapsed adjustment (#215 C6, C7)

erikdarlingdata:
## Summary
Joe's feedback from #215, public plan \`MwX36LEHJS\`:

- **C6**: the current parallel-wait formula uses per-thread \`elapsed - cpu\`, which is ≈0 for external/preemptive waits where the worker is CPU-busy in kernel (MEMORY_ALLOCATION_EXT, RESERVED_MEMORY_ALLOCATION_EXT, PREEMPTIVE_*). New path computes \`(wait_ms / total_cpu_ms) * Σ max-thread-self-cpu / elapsed_ms\`.
- **C7**: CPU:Elapsed ratio subtracts Σ external-wait ms from CPU before dividing — those waits aren't real query CPU.

## Verified
- Joe's worked example: \`(26307/45075) * (11341 + 42) / 13748 = 48.3%\` — formula matches.
- Private plan \`c1-c5.sqlplan\` (serial): RESERVED_MEMORY_ALLOCATION_EXT benefit → 0.5% (serial: sum-of-self-cpu ≈ stmt cpu, so formula collapses to simple ratio — no regression).
- Regression on \`20260415_1.sqlplan\` (parallel batch-mode): CXPACKET 99.2%, ASYNC_NETWORK_IO 93.5% unchanged; MEMORY_ALLOCATION_EXT stays at 0.1%.

## Implementation notes
- \`BenefitScorer.IsExternalWait\` classifier (public — ResultMapper needs it).
- \`OperatorWaitProfile\` grows \`MaxThreadCpuMs\`, populated via new \`PlanAnalyzer.GetOperatorMaxThreadOwnCpuMs\` which mirrors the self-elapsed helper: per-thread self for parallel, operator-minus-children for serial. Crucial — using cumulative CPU would double-count in serial plans.
- \`QueryTimeResult\` gains \`ExternalWaitMs\` populated during mapping.
- CPU:Elapsed renderers updated in HtmlExporter runtime card, Index.razor runtime card, PlanViewerControl DOP efficiency calc.

## Test plan
- [x] Formula matches Joe's expected 48.3% by math
- [x] Private + public reference plans show expected numbers
- [x] Regression on existing parallel plan
- [ ] Verify against raw XML of MwX36LEHJS when available

Version 1.7.3 → 1.7.4.

### DIFF
--- a/src/PlanViewer.App/Controls/PlanViewerControl.axaml.cs
+++ b/src/PlanViewer.App/Controls/PlanViewerControl.axaml.cs
@@ -2839,8 +2839,15 @@ public partial class PlanViewerControl : UserControl
                 statement.QueryTimeStats.CpuTimeMs > 0 &&
                 statement.DegreeOfParallelism > 1)
             {
-                // Speedup ratio: CPU/elapsed = 1.0 means serial, = DOP means perfect parallelism
-                var speedup = (double)statement.QueryTimeStats.CpuTimeMs / statement.QueryTimeStats.ElapsedTimeMs;
+                // Speedup ratio: CPU/elapsed = 1.0 means serial, = DOP means perfect parallelism.
+                // Subtract external/preemptive wait time from CPU — those waits are CPU-busy
+                // in kernel and inflate the ratio without representing real query work.
+                long externalWaitMs = 0;
+                foreach (var w in statement.WaitStats)
+                    if (BenefitScorer.IsExternalWait(w.WaitType))
+                        externalWaitMs += w.WaitTimeMs;
+                var effectiveCpu = Math.Max(0, statement.QueryTimeStats.CpuTimeMs - externalWaitMs);
+                var speedup = (double)effectiveCpu / statement.QueryTimeStats.ElapsedTimeMs;
                 var efficiency = Math.Min(100.0, (speedup - 1.0) / (statement.DegreeOfParallelism - 1.0) * 100.0);
                 efficiency = Math.Max(0.0, efficiency);
                 dopText += $" ({efficiency:N0}% efficient)";
--- a/src/PlanViewer.Core/Output/ResultMapper.cs
+++ b/src/PlanViewer.Core/Output/ResultMapper.cs
@@ -1,4 +1,5 @@
 using PlanViewer.Core.Models;
+using PlanViewer.Core.Services;
 
 namespace PlanViewer.Core.Output;
 
@@ -111,10 +112,18 @@ public static class ResultMapper
         // Query time (actual plans)
         if (stmt.QueryTimeStats != null)
         {
+            long externalWaitMs = 0;
+            foreach (var w in stmt.WaitStats)
+            {
+                if (BenefitScorer.IsExternalWait(w.WaitType))
+                    externalWaitMs += w.WaitTimeMs;
+            }
+
             result.QueryTime = new QueryTimeResult
             {
                 CpuTimeMs = stmt.QueryTimeStats.CpuTimeMs,
-                ElapsedTimeMs = stmt.QueryTimeStats.ElapsedTimeMs
+                ElapsedTimeMs = stmt.QueryTimeStats.ElapsedTimeMs,
+                ExternalWaitMs = externalWaitMs
             };
         }
 
--- a/src/PlanViewer.Core/Services/BenefitScorer.cs
+++ b/src/PlanViewer.Core/Services/BenefitScorer.cs
@@ -494,12 +494,11 @@ public static class BenefitScorer
         var isParallel = stmt.DegreeOfParallelism > 1 && stmt.RootNode != null;
 
         // Collect all operators with per-thread stats for parallel benefit calculation
-        List<OperatorWaitProfile>? operatorProfiles = null;
-        if (isParallel)
-        {
-            operatorProfiles = new List<OperatorWaitProfile>();
-            CollectOperatorWaitProfiles(stmt.RootNode!, operatorProfiles);
-        }
+        // Collect operator profiles even for serial plans — the external-wait formula
+        // uses sum-of-max-thread-cpu across operators and works for both.
+        var operatorProfiles = new List<OperatorWaitProfile>();
+        if (stmt.RootNode != null)
+            CollectOperatorWaitProfiles(stmt.RootNode, operatorProfiles);
 
         foreach (var wait in stmt.WaitStats)
         {
@@ -508,7 +507,15 @@ public static class BenefitScorer
             var category = ClassifyWaitType(wait.WaitType);
             double benefitPct;
 
-            if (category == "Parallelism" && isParallel)
+            if (IsExternalWait(wait.WaitType) && operatorProfiles.Count > 0)
+            {
+                // External / preemptive waits (MEMORY_ALLOCATION_*, PREEMPTIVE_*): the worker
+                // is CPU-busy in kernel, so operator elapsed ≈ operator cpu and the wait
+                // barely shows in the per-thread (elapsed - cpu) calculation. Joe's formula:
+                //   benefit = (wait_ms / total_cpu_ms) * Σ max_thread_cpu_per_operator / elapsed
+                benefitPct = CalculateExternalWaitBenefit(wait, operatorProfiles, stmt.QueryTimeStats!.CpuTimeMs, elapsedMs);
+            }
+            else if (category == "Parallelism" && isParallel)
             {
                 // CXPACKET/CXCONSUMER/CXSYNC: benefit is the parallelism efficiency gap,
                 // not the raw wait time. Threads waiting for other threads is a symptom
@@ -525,7 +532,7 @@ public static class BenefitScorer
                     benefitPct = (double)wait.WaitTimeMs / elapsedMs * 100;
                 }
             }
-            else if (!isParallel || operatorProfiles == null || operatorProfiles.Count == 0)
+            else if (!isParallel || operatorProfiles.Count == 0)
             {
                 // Serial plan or no operator data: simple ratio
                 benefitPct = (double)wait.WaitTimeMs / elapsedMs * 100;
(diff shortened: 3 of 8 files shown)


src/PlanViewer.Core/Services/BenefitScorer.cs
```diff
@@ -585,6 +592,48 @@ private static double CalculateParallelWaitBenefit(
...
+    /// External / preemptive waits where the worker is CPU-busy in kernel rather than
+    /// descheduled. Their wait time counts toward the query's CPU time, so the usual
+    /// (elapsed - cpu) per-thread wait math misses them entirely.
+    /// </summary>
+    public static bool IsExternalWait(string waitType)
+    {
+        if (string.IsNullOrEmpty(waitType)) return false;
+        var wt = waitType.ToUpperInvariant();
+        return wt.Contains("MEMORY_ALLOCATION")
+            || wt.StartsWith("PREEMPTIVE_");
```
erikdarlingdata, line 634: `Contains("MEMORY_ALLOCATION")` catches both `MEMORY_ALLOCATION_EXT` and `RESERVED_MEMORY_ALLOCATION_EXT` — that's the intent — but it's substring-wide, so any future wait type embedding that token anywhere in the name would be (mis)classified as CPU-busy. Tighter alternative:

```csharp
return wt == "MEMORY_ALLOCATION_EXT"
    || wt == "RESERVED_MEMORY_ALLOCATION_EXT"
    || wt.StartsWith("PREEMPTIVE_");
```

Not a bug today, just a robustness nit. Same classifier is reused from `ResultMapper` and `PlanViewerControl`, so whatever shape it takes here is authoritative.

---

src/PlanViewer.Core/Services/PlanAnalyzer.cs
```diff
@@ -1592,6 +1592,66 @@ private static long GetPerThreadOwnElapsed(PlanNode node)
...
+            var parentByThread = new Dictionary<int, long>();
+            foreach (var ts in node.PerThreadStats)
+                parentByThread[ts.ThreadId] = ts.ActualCPUMs;
+
+            var childSumByThread = new Dictionary<int, long>();
+            foreach (var child in node.Children)
+            {
+                var childNode = child;
+                if (child.PhysicalOp == "Parallelism" && child.Children.Count > 0)
+                    childNode = child.Children.OrderByDescending(c => c.ActualCPUMs).First();
```
erikdarlingdata, line 1616: Exchange look-through mirrors `GetPerThreadOwnElapsed` at 1573-1574 — good, pattern is consistent. One subtle thing: on the parent side (`parentByThread`), thread IDs come from the operator itself; on the child side (`childSumByThread`), they come from the **grandchild** when the direct child is `Parallelism`. Producer-side threads below a parallelism exchange are a different thread set from consumer-side threads above it, so `childSumByThread` lookups keyed by parent's thread IDs will often miss and contribute 0. That's actually the right answer (you don't want to subtract producer CPU from consumer thread self-CPU), but it means in those cases this collapses to "parent thread cpu" as self. Worth a one-line comment noting that's intentional.

---

src/PlanViewer.App/Controls/PlanViewerControl.axaml.cs
```diff
@@ -2839,8 +2839,15 @@ static string EfficiencyColor(double pct) => pct >= 80 ? "#E4E6EB"
...
                 statement.DegreeOfParallelism > 1)
             {
-                // Speedup ratio: CPU/elapsed = 1.0 means serial, = DOP means perfect parallelism
-                var speedup = (double)statement.QueryTimeStats.CpuTimeMs / statement.QueryTimeStats.ElapsedTimeMs;
+                // Speedup ratio: CPU/elapsed = 1.0 means serial, = DOP means perfect parallelism.
+                // Subtract external/preemptive wait time from CPU — those waits are CPU-busy
+                // in kernel and inflate the ratio without representing real query work.
+                long externalWaitMs = 0;
+                foreach (var w in statement.WaitStats)
+                    if (BenefitScorer.IsExternalWait(w.WaitType))
```
erikdarlingdata, line 2847: `ResultMapper` already derives `QueryTimeResult.ExternalWaitMs` using the same classifier (`ResultMapper.cs:115-119`). This code-behind works against the raw `PlanStatement`, not the mapped `StatementResult`, so the re-sum is necessary — just flagging that the classifier (`BenefitScorer.IsExternalWait`) is the single source of truth across all three sites (here, `ResultMapper`, `HtmlExporter`/`Index.razor` via the mapped field). If the classifier ever drifts, this site would need to drift with it.

---

src/PlanViewer.Core/Services/BenefitScorer.cs
```diff
@@ -494,12 +494,11 @@ private static void ScoreWaitStats(PlanStatement stmt)
...
-        if (isParallel)
-        {
-            operatorProfiles = new List<OperatorWaitProfile>();
-            CollectOperatorWaitProfiles(stmt.RootNode!, operatorProfiles);
-        }
+        // Collect operator profiles even for serial plans — the external-wait formula
+        // uses sum-of-max-thread-cpu across operators and works for both.
+        var operatorProfiles = new List<OperatorWaitProfile>();
+        if (stmt.RootNode != null)
+            CollectOperatorWaitProfiles(stmt.RootNode, operatorProfiles);
```
erikdarlingdata, line 501: Previously the tree walk was skipped for serial plans. Now `CollectOperatorWaitProfiles` runs on every statement that has a root node, including serial plans that have no external waits. For statements with many operators and no `MEMORY_ALLOCATION_*`/`PREEMPTIVE_*` waits this is pure overhead. If it shows up in CLI/bulk-scan profiling, gate the walk on `stmt.WaitStats.Any(w => IsExternalWait(w.WaitType)) || isParallel`.

---

src/PlanViewer.Core/Services/BenefitScorer.cs
```diff
@@ -508,7 +507,15 @@ private static void ScoreWaitStats(PlanStatement stmt)
...
             double benefitPct;
 
-            if (category == "Parallelism" && isParallel)
+            if (IsExternalWait(wait.WaitType) && operatorProfiles.Count > 0)
+            {
+                // External / preemptive waits (MEMORY_ALLOCATION_*, PREEMPTIVE_*): the worker
+                // is CPU-busy in kernel, so operator elapsed ≈ operator cpu and the wait
+                // barely shows in the per-thread (elapsed - cpu) calculation. Joe's formula:
+                //   benefit = (wait_ms / total_cpu_ms) * Σ max_thread_cpu_per_operator / elapsed
+                benefitPct = CalculateExternalWaitBenefit(wait, operatorProfiles, stmt.QueryTimeStats!.CpuTimeMs, elapsedMs);
```
erikdarlingdata, line 516: The external-wait path is taken for serial plans too, not just parallel. On a serial plan with per-operator CPU stats, `sumMaxCpu ≈ stmt.CpuTimeMs` and the formula collapses to `wait_ms / elapsed_ms` — which is exactly the old serial behavior, so no regression (matches your c1-c5.sqlplan verification). Worth noting in the doc-comment above (`Σ max_thread_cpu_per_operator` reads as parallel-only at a glance).

---

src/PlanViewer.Core/Output/ResultMapper.cs
```diff
@@ -111,10 +112,18 @@ private static StatementResult MapStatement(PlanStatement stmt)
...
+                if (BenefitScorer.IsExternalWait(w.WaitType))
+                    externalWaitMs += w.WaitTimeMs;
+            }
+
             result.QueryTime = new QueryTimeResult
             {
                 CpuTimeMs = stmt.QueryTimeStats.CpuTimeMs,
-                ElapsedTimeMs = stmt.QueryTimeStats.ElapsedTimeMs
+                ElapsedTimeMs = stmt.QueryTimeStats.ElapsedTimeMs,
+                ExternalWaitMs = externalWaitMs
```
erikdarlingdata, line 126: No test coverage landing with this PR. `tests/PlanViewer.Core.Tests/PlanAnalyzerTests.cs` has wait-stat tests (Rule 11 severity gating, CPU vs I/O waits) but nothing asserting on benefit % values. A small fixture covering:

- serial plan with a `RESERVED_MEMORY_ALLOCATION_EXT` wait → asserts `ExternalWaitMs` is populated and benefit% is reasonable;
- a plan matching Joe's 48.3% worked example → locks that math in so a future refactor can't silently move the number.

Would catch any future drift in either `IsExternalWait` or the formula. Not blocking, but flagging per the review checklist.

---

erikdarlingdata (commented):
## What this PR does

Addresses #215 C6 and C7. External/preemptive waits (`MEMORY_ALLOCATION_*`, `PREEMPTIVE_*`) are CPU-busy in kernel, so the existing per-thread `elapsed − cpu` benefit formula scores them near zero. This PR:

- Adds `BenefitScorer.IsExternalWait` and a dedicated `CalculateExternalWaitBenefit` using Joe's `(wait_ms / cpu_ms) * Σ max-thread-self-cpu / elapsed_ms`.
- Adds `PlanAnalyzer.GetOperatorMaxThreadOwnCpuMs` (the CPU analogue of `GetPerThreadOwnElapsed`) to feed the numerator.
- Surfaces `QueryTimeResult.ExternalWaitMs` and subtracts it from CPU in the CPU:Elapsed ratio across `HtmlExporter`, `Index.razor`, and the DOP-efficiency calc in `PlanViewerControl`.
- Bumps version 1.7.3 → 1.7.4. Base branch is `dev`, correct.

## What's good

- Formula is derivation-consistent with the `(elapsed - cpu)` path for the non-external case; serial plans collapse to the old simple ratio so there's no regression path (confirmed by your c1-c5 verification).
- Single classifier (`IsExternalWait`) used in all four consumers — no drift risk as long as it stays authoritative.
- `GetOperatorMaxThreadOwnCpuMs` mirrors `GetPerThreadOwnElapsed`'s exchange look-through pattern; serial branch clamps with `Max(0, ...)` for the known pass-through / timing-skew cases.

## Needs attention

1. **Cross-repo rule-set drift.** This modifies `BenefitScorer.ScoreWaitStats` and adds a new `PlanAnalyzer` helper — both are part of the analysis rule set shared with PerformanceMonitor's Dashboard and Lite copies. The PR body doesn't mention sync'ing them. If those are meant to stay in lockstep with Studio, companion PRs need to land there, otherwise the same plan will score differently in each product.

2. **No tests.** `tests/PlanViewer.Core.Tests/` has nothing asserting on benefit percentages for external waits, and nothing that would catch a regression in Joe's 48.3% worked example. Inline comment on `ResultMapper.cs` has suggested coverage.

3. **Minor, inline:**
   - `IsExternalWait` uses `Contains("MEMORY_ALLOCATION")` — works, but broader than needed.
   - `CollectOperatorWaitProfiles` now walks for every statement with a root node, not just parallel — cheap per-plan but adds up in CLI bulk runs; gate-on-need is straightforward if it matters.
   - Doc-comment on `CalculateExternalWaitBenefit` reads parallel-only, but the function is intentionally also the serial path — worth a line noting that.

4. **Open TODOs in the PR body** (raw XML for `MwX36LEHJS`, web visual post-deploy) — not blocking review, just flagging they're still unchecked.

No Avalonia concerns (no new controls, no XAML styles, no MVVM creep, no contrast-range brushes introduced).

---

PR #260, opened 2026-04-22
Release v1.7.4 — external-wait formula + CPU:Elapsed adjustment

erikdarlingdata:
## Summary
Joe feedback #215 C6 + C7:

- External/preemptive waits (MEMORY_ALLOCATION_EXT, RESERVED_MEMORY_ALLOCATION_EXT, PREEMPTIVE_*) now use the CPU-share formula so benefit % matches query impact instead of being near-zero.
- CPU:Elapsed ratio subtracts external-wait time from CPU for a truer picture.

See PR #259.

## Test plan
- [x] Formula math matches Joe's expected 48.3%
- [x] Serial and parallel reference plans sanity-check
- [ ] Visual post-deploy
- [ ] Verify MwX36LEHJS raw XML when received

### DIFF
--- a/src/PlanViewer.App/Controls/PlanViewerControl.axaml.cs
+++ b/src/PlanViewer.App/Controls/PlanViewerControl.axaml.cs
@@ -2839,8 +2839,15 @@ public partial class PlanViewerControl : UserControl
                 statement.QueryTimeStats.CpuTimeMs > 0 &&
                 statement.DegreeOfParallelism > 1)
             {
-                // Speedup ratio: CPU/elapsed = 1.0 means serial, = DOP means perfect parallelism
-                var speedup = (double)statement.QueryTimeStats.CpuTimeMs / statement.QueryTimeStats.ElapsedTimeMs;
+                // Speedup ratio: CPU/elapsed = 1.0 means serial, = DOP means perfect parallelism.
+                // Subtract external/preemptive wait time from CPU — those waits are CPU-busy
+                // in kernel and inflate the ratio without representing real query work.
+                long externalWaitMs = 0;
+                foreach (var w in statement.WaitStats)
+                    if (BenefitScorer.IsExternalWait(w.WaitType))
+                        externalWaitMs += w.WaitTimeMs;
+                var effectiveCpu = Math.Max(0, statement.QueryTimeStats.CpuTimeMs - externalWaitMs);
+                var speedup = (double)effectiveCpu / statement.QueryTimeStats.ElapsedTimeMs;
                 var efficiency = Math.Min(100.0, (speedup - 1.0) / (statement.DegreeOfParallelism - 1.0) * 100.0);
                 efficiency = Math.Max(0.0, efficiency);
                 dopText += $" ({efficiency:N0}% efficient)";
--- a/src/PlanViewer.Core/Output/AnalysisResult.cs
+++ b/src/PlanViewer.Core/Output/AnalysisResult.cs
@@ -176,6 +176,14 @@ public class QueryTimeResult
 
     [JsonPropertyName("elapsed_time_ms")]
     public long ElapsedTimeMs { get; set; }
+
+    /// <summary>
+    /// Sum of external/preemptive wait time (MEMORY_ALLOCATION_*, PREEMPTIVE_*) —
+    /// these waits are CPU-busy in kernel and inflate CpuTimeMs vs real query CPU.
+    /// Subtract from CpuTimeMs for a truer CPU:Elapsed ratio.
+    /// </summary>
+    [JsonPropertyName("external_wait_ms")]
+    public long ExternalWaitMs { get; set; }
 }
 
 public class ParameterResult
--- a/src/PlanViewer.Core/Output/ResultMapper.cs
+++ b/src/PlanViewer.Core/Output/ResultMapper.cs
@@ -1,4 +1,5 @@
 using PlanViewer.Core.Models;
+using PlanViewer.Core.Services;
 
 namespace PlanViewer.Core.Output;
 
@@ -111,10 +112,18 @@ public static class ResultMapper
         // Query time (actual plans)
         if (stmt.QueryTimeStats != null)
         {
+            long externalWaitMs = 0;
+            foreach (var w in stmt.WaitStats)
+            {
+                if (BenefitScorer.IsExternalWait(w.WaitType))
+                    externalWaitMs += w.WaitTimeMs;
+            }
+
             result.QueryTime = new QueryTimeResult
             {
                 CpuTimeMs = stmt.QueryTimeStats.CpuTimeMs,
-                ElapsedTimeMs = stmt.QueryTimeStats.ElapsedTimeMs
+                ElapsedTimeMs = stmt.QueryTimeStats.ElapsedTimeMs,
+                ExternalWaitMs = externalWaitMs
             };
         }
 
--- a/src/PlanViewer.Core/Services/BenefitScorer.cs
+++ b/src/PlanViewer.Core/Services/BenefitScorer.cs
@@ -494,12 +494,11 @@ public static class BenefitScorer
         var isParallel = stmt.DegreeOfParallelism > 1 && stmt.RootNode != null;
 
         // Collect all operators with per-thread stats for parallel benefit calculation
-        List<OperatorWaitProfile>? operatorProfiles = null;
-        if (isParallel)
-        {
-            operatorProfiles = new List<OperatorWaitProfile>();
-            CollectOperatorWaitProfiles(stmt.RootNode!, operatorProfiles);
-        }
+        // Collect operator profiles even for serial plans — the external-wait formula
+        // uses sum-of-max-thread-cpu across operators and works for both.
+        var operatorProfiles = new List<OperatorWaitProfile>();
+        if (stmt.RootNode != null)
+            CollectOperatorWaitProfiles(stmt.RootNode, operatorProfiles);
 
         foreach (var wait in stmt.WaitStats)
         {
@@ -508,7 +507,15 @@ public static class BenefitScorer
             var category = ClassifyWaitType(wait.WaitType);
             double benefitPct;
 
-            if (category == "Parallelism" && isParallel)
+            if (IsExternalWait(wait.WaitType) && operatorProfiles.Count > 0)
+            {
+                // External / preemptive waits (MEMORY_ALLOCATION_*, PREEMPTIVE_*): the worker
+                // is CPU-busy in kernel, so operator elapsed ≈ operator cpu and the wait
+                // barely shows in the per-thread (elapsed - cpu) calculation. Joe's formula:
+                //   benefit = (wait_ms / total_cpu_ms) * Σ max_thread_cpu_per_operator / elapsed
+                benefitPct = CalculateExternalWaitBenefit(wait, operatorProfiles, stmt.QueryTimeStats!.CpuTimeMs, elapsedMs);
+            }
+            else if (category == "Parallelism" && isParallel)
             {
                 // CXPACKET/CXCONSUMER/CXSYNC: benefit is the parallelism efficiency gap,
                 // not the raw wait time. Threads waiting for other threads is a symptom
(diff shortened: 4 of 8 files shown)


src/PlanViewer.App/Controls/PlanViewerControl.axaml.cs
```diff
@@ -2839,8 +2839,15 @@ static string EfficiencyColor(double pct) => pct >= 80 ? "#E4E6EB"
...
-                // Speedup ratio: CPU/elapsed = 1.0 means serial, = DOP means perfect parallelism
-                var speedup = (double)statement.QueryTimeStats.CpuTimeMs / statement.QueryTimeStats.ElapsedTimeMs;
+                // Speedup ratio: CPU/elapsed = 1.0 means serial, = DOP means perfect parallelism.
+                // Subtract external/preemptive wait time from CPU — those waits are CPU-busy
+                // in kernel and inflate the ratio without representing real query work.
+                long externalWaitMs = 0;
+                foreach (var w in statement.WaitStats)
+                    if (BenefitScorer.IsExternalWait(w.WaitType))
+                        externalWaitMs += w.WaitTimeMs;
+                var effectiveCpu = Math.Max(0, statement.QueryTimeStats.CpuTimeMs - externalWaitMs);
```
erikdarlingdata, line 2849: Third copy of the external-wait summation (also in `ResultMapper.cs:114-119` and implicit in `BenefitScorer.CalculateExternalWaitBenefit`). Three call sites iterate `WaitStats`, call `IsExternalWait`, sum `WaitTimeMs`. Worth pulling into a single helper on `PlanStatement` (or `BenefitScorer.SumExternalWaitMs(stmt)`) so the definition of "external wait" has one home — especially if the rule ever expands beyond `MEMORY_ALLOCATION` / `PREEMPTIVE_*`.

---

src/PlanViewer.Core/Services/BenefitScorer.cs
```diff
@@ -585,6 +592,48 @@ private static double CalculateParallelWaitBenefit(
...
+    /// External / preemptive waits where the worker is CPU-busy in kernel rather than
+    /// descheduled. Their wait time counts toward the query's CPU time, so the usual
+    /// (elapsed - cpu) per-thread wait math misses them entirely.
+    /// </summary>
+    public static bool IsExternalWait(string waitType)
+    {
+        if (string.IsNullOrEmpty(waitType)) return false;
+        var wt = waitType.ToUpperInvariant();
+        return wt.Contains("MEMORY_ALLOCATION")
+            || wt.StartsWith("PREEMPTIVE_");
```
erikdarlingdata, line 634: `Contains("MEMORY_ALLOCATION")` is broader than the doc-comment's stated target (`MEMORY_ALLOCATION_EXT`, `RESERVED_MEMORY_ALLOCATION_EXT`). Any future wait type with that substring will silently be classified as external and get the CPU-share formula even if the worker isn't CPU-busy in kernel. Tightening to `StartsWith("MEMORY_ALLOCATION") || StartsWith("RESERVED_MEMORY_ALLOCATION")` — or an explicit allowlist — matches the documented intent.

Also note this diverges from `ClassifyWaitType` at line 550, which still puts `MEMORY_ALLOCATION` in the `"Memory"` category. Two classifiers of the same wait that disagree is a future foot-gun.

---

src/PlanViewer.Core/Services/BenefitScorer.cs
```diff
@@ -585,6 +592,48 @@ private static double CalculateParallelWaitBenefit(
...
+    ///   sum_max_cpu = Σ max_thread_cpu across operators
+    ///   benefit_ms  = wait_cpu_share * sum_max_cpu
+    /// Then convert to % of statement elapsed.
+    /// </summary>
+    private static double CalculateExternalWaitBenefit(
+        WaitStatInfo wait, List<OperatorWaitProfile> profiles,
+        long stmtCpuMs, long stmtElapsedMs)
+    {
+        if (stmtCpuMs <= 0 || stmtElapsedMs <= 0)
+            return (double)wait.WaitTimeMs / Math.Max(1, stmtElapsedMs) * 100;
```
erikdarlingdata, line 610: Early-return path when `stmtCpuMs <= 0` falls back to `wait_ms / elapsed * 100` — the same math that was already failing to surface these waits (per the method's own doc-comment). If the data is missing, returning 0 would be more honest than a known-wrong number.

---

src/PlanViewer.Core/Services/PlanAnalyzer.cs
```diff
@@ -1592,6 +1592,66 @@ private static long GetPerThreadOwnElapsed(PlanNode node)
...
+            var parentByThread = new Dictionary<int, long>();
+            foreach (var ts in node.PerThreadStats)
+                parentByThread[ts.ThreadId] = ts.ActualCPUMs;
+
+            var childSumByThread = new Dictionary<int, long>();
+            foreach (var child in node.Children)
+            {
+                var childNode = child;
+                if (child.PhysicalOp == "Parallelism" && child.Children.Count > 0)
+                    childNode = child.Children.OrderByDescending(c => c.ActualCPUMs).First();
```
erikdarlingdata, line 1616: `OrderByDescending(c => c.ActualCPUMs).First()` picks a single Parallelism-exchange child and discards the other side(s), so any CPU those threads contributed is dropped from the per-thread subtraction. On ties, `.First()` is effectively arbitrary (enumeration order) which makes `maxSelf` non-deterministic for identical plan inputs.

Two questions worth confirming: (1) Is picking the max-CPU child actually the right semantic here, versus summing per-thread CPU across all exchange children keyed by ThreadId? (2) Does this mirror the existing elapsed-path helper `GetPerThreadOwnElapsed` — if not, the two "critical-path" calculations may diverge.

---

src/PlanViewer.Core/Services/PlanAnalyzer.cs
```diff
@@ -1592,6 +1592,66 @@ private static long GetPerThreadOwnElapsed(PlanNode node)
...
+    /// </summary>
+    internal static long GetOperatorMaxThreadOwnCpuMs(PlanNode node)
+    {
+        if (!node.HasActualStats || node.ActualCPUMs <= 0) return 0;
+
+        if (node.PerThreadStats.Count > 1)
+        {
+            var parentByThread = new Dictionary<int, long>();
+            foreach (var ts in node.PerThreadStats)
+                parentByThread[ts.ThreadId] = ts.ActualCPUMs;
```
erikdarlingdata, line 1609: `parentByThread[ts.ThreadId] = ts.ActualCPUMs` — if `PerThreadStats` ever contains two entries for the same `ThreadId` (e.g., thread reuse across iterations, or parser quirks on some plan shapes), the second silently overwrites the first. Worth a `TryGetValue + sum` like `childSumByThread` uses a few lines below, for symmetry and safety.

---

src/PlanViewer.Core/Services/BenefitScorer.cs
```diff
@@ -508,7 +507,15 @@ private static void ScoreWaitStats(PlanStatement stmt)
             var category = ClassifyWaitType(wait.WaitType);
             double benefitPct;
```
erikdarlingdata, line 508: Gate `IsExternalWait(...) && operatorProfiles.Count > 0`: for an external wait on a plan without a `RootNode` (estimated-only plan that somehow has wait stats, or parse failures), this falls through to the `Parallelism` branch test and then to `!isParallel || operatorProfiles.Count == 0` → simple ratio. That's the pre-fix behavior, which the doc-comments say is wrong for these waits. Likely fine in practice, but the fallback chain for external waits in degraded data is worth a single explicit branch.

---

src/PlanViewer.Core/Output/ResultMapper.cs
```diff
@@ -1,4 +1,5 @@
 using PlanViewer.Core.Models;
+using PlanViewer.Core.Services;
```
erikdarlingdata, line 2: Output layer now depends on `PlanViewer.Core.Services` to call `BenefitScorer.IsExternalWait`. Fine as long as you're OK with Output → Services, but the classification is really a model concern — a `PlanStatement.ExternalWaitMs` property (or a `WaitStatInfo.IsExternal` computed once during parse) would let both Output and the Avalonia control consume the value without reaching into Services.

---

src/PlanViewer.Core/Output/AnalysisResult.cs
```diff
@@ -176,6 +176,14 @@ public class QueryTimeResult
...
     [JsonPropertyName("elapsed_time_ms")]
     public long ElapsedTimeMs { get; set; }
+
+    /// <summary>
+    /// Sum of external/preemptive wait time (MEMORY_ALLOCATION_*, PREEMPTIVE_*) —
+    /// these waits are CPU-busy in kernel and inflate CpuTimeMs vs real query CPU.
+    /// Subtract from CpuTimeMs for a truer CPU:Elapsed ratio.
+    /// </summary>
+    [JsonPropertyName("external_wait_ms")]
+    public long ExternalWaitMs { get; set; }
```
erikdarlingdata, line 186: New JSON field `external_wait_ms` is a public output contract — MCP clients, downstream consumers, and the Dashboard/Lite copies of PerformanceMonitor will start seeing this. Worth mentioning in release notes and confirming any consumer that computes its own CPU:Elapsed ratio (e.g., dashboards) gets the same treatment.

---

erikdarlingdata (commented):
## Summary
v1.7.4 release: external/preemptive waits (`MEMORY_ALLOCATION_*`, `PREEMPTIVE_*`) now get CPU-share benefit scoring instead of the `(elapsed - cpu)` per-thread math that missed them, and CPU:Elapsed ratios subtract external-wait time from CPU. New `QueryTimeResult.ExternalWaitMs` field exposes the delta. Adds `PlanAnalyzer.GetOperatorMaxThreadOwnCpuMs` to compute per-thread self-CPU for the new formula.

## What's good
- Targeted fix with a specific math rationale (`wait_cpu_share * Σ max_thread_cpu`) and doc-comments explaining the why.
- `CpuTimeMs - ExternalWaitMs` is clamped with `Math.Max(0, ...)` at every call site, so a bad wait-stats parse can't produce a negative ratio.
- `ScoreWaitStats` now collects `operatorProfiles` unconditionally (not just for parallel), which the new formula needs.

## What needs attention

### Base branch
**PR targets `main`, not `dev`.** Release PRs or not, that's contrary to the project's branching convention — confirm this is intentional before merging.

### Cross-repo sync (PerformanceMonitor)
This touches the plan-analysis rule set (`BenefitScorer.ScoreWaitStats`, wait classification, external-wait formula) and the `PlanAnalyzer` API surface (`GetOperatorMaxThreadOwnCpuMs`). The equivalent change must also land in the PerformanceMonitor **Dashboard** and **Lite** copies, or their rule set will drift from Studio's on every external-wait plan. Track separately.

### Tests
No changes under `tests/PlanViewer.Core.Tests/`. Logic added in `BenefitScorer.CalculateExternalWaitBenefit`, `BenefitScorer.IsExternalWait`, and `PlanAnalyzer.GetOperatorMaxThreadOwnCpuMs` has no unit coverage. At minimum:
- A parallel plan with a `MEMORY_ALLOCATION_EXT` wait and known per-thread CPU → asserts Joe's 48.3% (test plan references it).
- A serial plan with `PREEMPTIVE_*` wait → asserts serial branch of `GetOperatorMaxThreadOwnCpuMs`.
- `IsExternalWait` positive/negative cases, including the `RESERVED_MEMORY_ALLOCATION_EXT` substring-match variant.

The PR body's "Verify MwX36LEHJS raw XML when received" box is still unchecked; a captured fixture for that plan under `tests/PlanViewer.Core.Tests/Plans/` would be the natural home.

### Inline items
See line comments for: (1) triplicated external-wait summation across `PlanViewerControl.axaml.cs:2845`, `ResultMapper.cs:114`, and `BenefitScorer.CalculateExternalWaitBenefit`; (2) `IsExternalWait` `Contains("MEMORY_ALLOCATION")` being broader than the documented target and conflicting with `ClassifyWaitType`'s `"Memory"` bucket; (3) `OrderByDescending(...).First()` picking a single Parallelism-exchange child nondeterministically in `GetOperatorMaxThreadOwnCpuMs`; (4) `parentByThread` overwrite on duplicate ThreadId; (5) fallback-to-known-wrong-ratio when `stmtCpuMs <= 0`; (6) Output → Services dependency in `ResultMapper`.

No Avalonia-specific issues (no AvaloniaEdit/ComboBox/`:pointerover`/brush changes). No contrast regressions.

---

PR #261, opened 2026-04-22
Loosen memory grant colors + compile time as plan-level property (#215 C1, C4)

erikdarlingdata:
## Summary
- **C1**: memory grant color scale was too harsh (61% utilized showed orange). New tiers: ≥40% good, 20–39% warn, <20% bad. Operators spill near their max, so moderate utilization is fine; what we flag now is real over-granting.
- **C4**: compile time always visible in the Runtime panel (category B plan-level property), not just when Rule 19 fires above 1000ms.

## Details
- \`PlanViewerControl.EfficiencyColor\` loosened — affects memory grant, DOP efficiency, thread utilization consistently.
- \`HtmlExporter\` memory card eff class + \`Index.razor\` eff class updated.
- \`Compile: Nms\` row added to Runtime card on all three surfaces.
- Version 1.7.4 → 1.7.5.

## Test plan
- [x] Build green
- [ ] Visual: 61% grant now reads green; compile row visible on plans with >0 compile time

### DIFF
--- a/src/PlanViewer.App/Controls/PlanViewerControl.axaml.cs
+++ b/src/PlanViewer.App/Controls/PlanViewerControl.axaml.cs
@@ -2800,9 +2800,12 @@ public partial class PlanViewerControl : UserControl
             rowIndex++;
         }
 
-        // Efficiency thresholds: white >= 80%, yellow >= 60%, orange >= 40%, red < 40%
-        static string EfficiencyColor(double pct) => pct >= 80 ? "#E4E6EB"
-            : pct >= 60 ? "#FFD700" : pct >= 40 ? "#FFB347" : "#E57373";
+        // Efficiency thresholds: white >= 40%, orange >= 20%, red < 20%.
+        // Loosened per Joe's feedback (#215 C1): for memory grants, moderate
+        // utilization (e.g. 60%) is fine — operators can spill near their max,
+        // so we shouldn't flag anything above a real over-grant threshold.
+        static string EfficiencyColor(double pct) => pct >= 40 ? "#E4E6EB"
+            : pct >= 20 ? "#FFB347" : "#E57373";
 
         // Runtime stats (actual plans)
         if (statement.QueryTimeStats != null)
@@ -2814,6 +2817,11 @@ public partial class PlanViewerControl : UserControl
             if (statement.QueryUdfElapsedTimeMs > 0)
                 AddRow("UDF elapsed", $"{statement.QueryUdfElapsedTimeMs:N0}ms");
         }
+
+        // Compile time — plan-level property (category B). Show regardless of
+        // threshold so it's always visible, not just when Rule 19 fires.
+        if (statement.CompileTimeMs > 0)
+            AddRow("Compile", $"{statement.CompileTimeMs:N0}ms");
 
         // Memory grant — color by utilization percentage
         if (statement.MemoryGrant != null)
--- a/src/PlanViewer.Core/Output/HtmlExporter.cs
+++ b/src/PlanViewer.Core/Output/HtmlExporter.cs
@@ -308,6 +308,8 @@ pre.query-text, pre.text-output {
                 WriteRow(sb, "CPU:Elapsed", ratio.ToString("N2"));
             }
         }
+        if (stmt.CompileTimeMs > 0)
+            WriteRow(sb, "Compile", $"{stmt.CompileTimeMs:N0} ms");
         if (stmt.DegreeOfParallelism > 0)
             WriteRow(sb, "DOP", stmt.DegreeOfParallelism.ToString());
         if (stmt.NonParallelReason != null)
@@ -315,7 +317,7 @@ pre.query-text, pre.text-output {
         if (stmt.MemoryGrant != null && stmt.MemoryGrant.GrantedKB > 0)
         {
             var pctUsed = (double)stmt.MemoryGrant.MaxUsedKB / stmt.MemoryGrant.GrantedKB * 100;
-            var effClass = pctUsed >= 80 ? "eff-good" : pctUsed >= 40 ? "eff-warn" : "eff-bad";
+            var effClass = pctUsed >= 40 ? "eff-good" : pctUsed >= 20 ? "eff-warn" : "eff-bad";
             WriteRow(sb, "Memory", FormatKB(stmt.MemoryGrant.GrantedKB) + " granted");
             sb.AppendLine($"<div class=\"row\"><span class=\"label\">Used</span><span class=\"value {effClass}\">{FormatKB(stmt.MemoryGrant.MaxUsedKB)} ({pctUsed:N0}%)</span></div>");
         }
--- a/src/PlanViewer.Web/Pages/Index.razor
+++ b/src/PlanViewer.Web/Pages/Index.razor
@@ -171,6 +171,13 @@ else
                         </div>
                     }
                 }
+                @if (ActiveStmt!.CompileTimeMs > 0)
+                {
+                    <div class="insight-row">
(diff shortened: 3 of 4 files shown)


src/PlanViewer.App/Controls/PlanViewerControl.axaml.cs
```diff
@@ -2815,6 +2818,11 @@ static string EfficiencyColor(double pct) => pct >= 80 ? "#E4E6EB"
                 AddRow("UDF elapsed", $"{statement.QueryUdfElapsedTimeMs:N0}ms");
         }
 
+        // Compile time — plan-level property (category B). Show regardless of
+        // threshold so it's always visible, not just when Rule 19 fires.
+        if (statement.CompileTimeMs > 0)
+            AddRow("Compile", $"{statement.CompileTimeMs:N0}ms");
```
erikdarlingdata, line 2824: Duplicate row. The same `BuildRuntimeSummary` method already emits compile time unconditionally at line 2895-2896:

```csharp
// Compile stats (always available)
if (statement.CompileTimeMs > 0)
    AddRow("Compile time", $"{statement.CompileTimeMs:N0}ms");
```

With this change, the Runtime panel will render two near-identical rows — `Compile` near the top and `Compile time` near the bottom. The rationale in the comment above ("Show regardless of threshold so it's always visible, not just when Rule 19 fires") doesn't match the existing code — this panel never gated compile time on Rule 19 firing. Pick one placement and delete the other; presumably the intent is to keep this new early placement next to the Runtime stats block and remove line 2895-2896.

---

src/PlanViewer.App/Controls/PlanViewerControl.axaml.cs
```diff
@@ -2800,9 +2800,12 @@ void AddRow(string label, string value, string? color = null)
...
 
-        // Efficiency thresholds: white >= 80%, yellow >= 60%, orange >= 40%, red < 40%
-        static string EfficiencyColor(double pct) => pct >= 80 ? "#E4E6EB"
-            : pct >= 60 ? "#FFD700" : pct >= 40 ? "#FFB347" : "#E57373";
+        // Efficiency thresholds: white >= 40%, orange >= 20%, red < 20%.
+        // Loosened per Joe's feedback (#215 C1): for memory grants, moderate
+        // utilization (e.g. 60%) is fine — operators can spill near their max,
+        // so we shouldn't flag anything above a real over-grant threshold.
+        static string EfficiencyColor(double pct) => pct >= 40 ? "#E4E6EB"
+            : pct >= 20 ? "#FFB347" : "#E57373";
```
erikdarlingdata, line 2808: `EfficiencyColor` is also called for DOP parallel efficiency (line 2862) and thread utilization (line 2878), so this threshold shift isn't memory-grant-only. Under the old scheme a plan running at ~50% parallel efficiency was yellow; now it's white/"good". Same for thread utilization. If that's intentional, the comment should say so; if not, you'll want a separate threshold for the parallelism/thread callers. Worth confirming before merge since the PR description frames this as a memory-grant tweak.

---

src/PlanViewer.Web/Pages/Index.razor
```diff
@@ -188,7 +195,7 @@ else
                 @if (ActiveStmt!.MemoryGrant != null && ActiveStmt!.MemoryGrant.GrantedKB > 0)
                 {
                     var pctUsed = (double)ActiveStmt!.MemoryGrant.MaxUsedKB / ActiveStmt!.MemoryGrant.GrantedKB * 100;
-                    var effClass = pctUsed >= 80 ? "eff-good" : pctUsed >= 60 ? "eff-ok" : pctUsed >= 40 ? "eff-warn" : "eff-bad";
+                    var effClass = pctUsed >= 40 ? "eff-good" : pctUsed >= 20 ? "eff-warn" : "eff-bad";
```
erikdarlingdata, line 198: The `eff-ok` tier is dropped here but `.eff-ok { color: #f9a825; }` still lives in `src/PlanViewer.Web/wwwroot/css/app.css:618`. No references remain after this change — safe to delete the CSS rule in the same PR to keep the three surfaces aligned.

---

src/PlanViewer.Core/Output/HtmlExporter.cs
```diff
@@ -308,14 +308,16 @@ private static void WriteRuntimeCard(StringBuilder sb, StatementResult stmt)
...
+            WriteRow(sb, "Compile", $"{stmt.CompileTimeMs:N0} ms");
         if (stmt.DegreeOfParallelism > 0)
             WriteRow(sb, "DOP", stmt.DegreeOfParallelism.ToString());
         if (stmt.NonParallelReason != null)
             WriteRow(sb, "Serial", Encode(stmt.NonParallelReason));
         if (stmt.MemoryGrant != null && stmt.MemoryGrant.GrantedKB > 0)
         {
             var pctUsed = (double)stmt.MemoryGrant.MaxUsedKB / stmt.MemoryGrant.GrantedKB * 100;
-            var effClass = pctUsed >= 80 ? "eff-good" : pctUsed >= 40 ? "eff-warn" : "eff-bad";
+            var effClass = pctUsed >= 40 ? "eff-good" : pctUsed >= 20 ? "eff-warn" : "eff-bad";
```
erikdarlingdata, line 320: Threshold shift here is isolated to the memory grant card, so this one reads cleanly — just noting for the record that `HtmlExporter` only applies the new thresholds to memory (unlike the App control, where the shared `EfficiencyColor` also affects DOP/thread). Three surfaces now behave slightly differently for parallelism coloring (App uses 40/20, HTML doesn't color DOP, Web doesn't color DOP). Probably fine but worth being explicit about.

---

erikdarlingdata (commented):
## Review summary

**What it does**
- Loosens the efficiency color ramp: good `>=80%` / warn `>=60%` / orange `>=40%` / bad becomes good `>=40%` / orange `>=20%` / bad across three surfaces (App `PlanViewerControl`, `HtmlExporter`, Web `Index.razor`).
- Adds a `Compile` row to the Runtime panel / card on all three surfaces so compile time shows whenever `CompileTimeMs > 0`.
- Bumps version 1.7.4 → 1.7.5.

**Good**
- Base branch is `dev`.
- New colors (`#E4E6EB`, `#FFB347`, `#E57373`) stay clear of the rejected dim-grey range.
- Pure code-behind, no MVVM creep.
- No PlanAnalyzer rule/scoring changes, so PerformanceMonitor Dashboard/Lite don't need a sibling sync for this PR.

**Needs attention**
1. **Duplicate compile row** in `PlanViewer.App/Controls/PlanViewerControl.axaml.cs` — the new `AddRow("Compile", …)` at 2823-2824 sits in the same `BuildRuntimeSummary` method as the existing `AddRow("Compile time", …)` at 2895-2896. Users will see two near-identical rows. Delete one. (See inline.)
2. **Shared `EfficiencyColor` side effects** at `PlanViewerControl.axaml.cs:2803-2808` — the loosened thresholds also apply to DOP parallel efficiency (2862) and thread utilization (2878), not just memory grants. PR description frames this as memory-only; confirm the parallelism/thread coloring change is intended. (See inline.)
3. **Dead CSS** — `.eff-ok` in `src/PlanViewer.Web/wwwroot/css/app.css:618` has no remaining references after this PR. Remove in the same change to keep surfaces consistent.
4. **No test updates** — thresholds are display-only so it's borderline, but `tests/PlanViewer.Core.Tests/` already covers memory grant rules (e.g. `Rule09a_*`, `LargeMemoryGrant_*`). A cheap regression test against the new ramp in `HtmlExporterTests.cs` would lock the 40/20 boundary in.

---

PR #302, opened 2026-04-29
Feature/allow application intent read only

ClaudioESSilva:
## What does this PR do?

Implements #301 

## Which component(s) does this affect?

- [x] Desktop App (PlanViewer.App)
- [x] Core Library (PlanViewer.Core)
- [ ] CLI Tool (PlanViewer.Cli)
- [ ] SSMS Extension (PlanViewer.Ssms)
- [ ] Tests
- [ ] Documentation

## How was this tested?

The "Connect to Server" dialog:
image

When connected to a Secondary, visually mention that.
image

## Checklist

- [x] I have read the [contributing guide](https://github.com/erikdarlingdata/PerformanceStudio/blob/main/CONTRIBUTING.md)
- [x] My code builds with zero warnings (`dotnet build -c Debug`)
- [x] All tests pass (`dotnet test`)
- [x] I have not introduced any hardcoded credentials or server names

PR #303, opened 2026-04-29
Connection dialog cleanup (post-#302)

erikdarlingdata:
Follow-up cleanup after #302 (read-only intent feature by @ClaudioESSilva). No behavior change — just polish.

## Changes

- **Rename `(Secondary)` indicator to `(Read-only)`** in `QuerySessionControl`. `ApplicationIntent=ReadOnly` *requests* secondary routing but doesn't guarantee it — the AG listener will still send the session to the primary if no readable replica is available. The label now reflects what we actually know (the user asked for read-only intent), not an assumption about topology.
- **Deduplicate `SqlConnectionStringBuilder` construction.** Previously the encrypt/trust/intent/auth logic was written twice — once in `ServerConnection.GetConnectionString(ICredentialService, ...)` for the runtime path, and once in `ConnectionDialog.BuildConnectionString` for the test path (which reads creds from the textboxes before they're saved). Added a `GetConnectionString(string? username, string? password, string? databaseName = null)` overload on `ServerConnection`; the credential-service overload now resolves creds and delegates to it, and the dialog calls it directly with textbox values. The dialog's `BuildConnectionString` is gone.
- **XAML nits** on the new Read-only intent checkbox: drop the redundant inner `Margin="0,0,0,6"` (the parent `StackPanel` already has `Margin="0,0,0,12"`) and trim trailing whitespace.

## Test plan

- [x] `dotnet build src/PlanViewer.Core` — clean
- [x] `dotnet build src/PlanViewer.App` — 4 pre-existing AVLN3001 warnings, no new ones
- [x] `dotnet build src/PlanViewer.Web` — clean
- [x] `dotnet test` — 75/75 pass
- [ ] Smoke-test in app: open ConnectionDialog, Test Connection still works for Windows, SQL Server, and Entra MFA auth modes
- [ ] Smoke-test in app: connect with Read-only intent checked, confirm `ServerLabel` shows `(Read-only)` after connect

### DIFF
--- a/src/PlanViewer.App/Controls/QuerySessionControl.axaml.cs
+++ b/src/PlanViewer.App/Controls/QuerySessionControl.axaml.cs
@@ -810,7 +810,7 @@ public partial class QuerySessionControl : UserControl
             _connectionString = _serverConnection.GetConnectionString(_credentialService, _selectedDatabase);
 
             ServerLabel.Text = _serverConnection.ApplicationIntentReadOnly
-                ? $"{_serverConnection.ServerName} (Secondary)"
+                ? $"{_serverConnection.ServerName} (Read-only)"
                 : _serverConnection.ServerName;
             ServerLabel.Foreground = Brushes.LimeGreen;
             ConnectButton.Content = "Reconnect";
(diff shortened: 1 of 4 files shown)


erikdarlingdata:
@ClaudioESSilva — small follow-up cleanup on top of your #302. No behavior change intended (the dialog test path now goes through the same `ServerConnection.GetConnectionString` that the runtime path uses, just with textbox creds passed in directly).

If you have a moment to smoke-test the Test Connection button across Windows / SQL Server / Entra MFA on your setup before this lands, that'd be appreciated since I don't have those auth modes available locally. No worries if not — happy to merge on green build + tests.

Issue #320, opened 2026-05-10
[FEATURE] Allow filter by "execution_type_desc" on Query Store screen


ClaudioESSilva:
### Which component(s) does this affect?

- [x] Desktop App
- [ ] CLI Tool
- [ ] SSMS Extension
- [ ] Plan Analysis Rules
- [ ] Documentation

### Problem Statement

When using [sp_QuickieStore](https://github.com/erikdarlingdata/DarlingData/blob/main/sp_QuickieStore/sp_QuickieStore.sql), I often use the filter `@execution_type_desc`.

I would like to have that possibility here too.

### Proposed Solution

1. Add a new option `Execution Type` to the `Search By` drop-down
2. Show a new drop-down `ExecutionTypeBox` with the three options: `regular`, `aborted` and `exception`, default to `regular`
3. This will apply an `AND rs.execution_type_desc = @executionTypeDesc`
4. If the user selects a different `Search By` option, the `ExecutionTypeBox` is hidden and the `SearchValueBox` text box is shown.
5. Add a column named "Execution Type" in both grids "QueryStoreGridControl" and "QueryStoreHistoryWindow".

### Use Case

Being able to filter Query Store entries related to execution type: `Aborted` or `Exception`.

### Alternatives Considered

Manually run [sp_QuickieStore](https://github.com/erikdarlingdata/DarlingData/blob/main/sp_QuickieStore/sp_QuickieStore.sql).

erikdarlingdata:
Implemented in #321 and merged to dev. CLI/MCP exposure followed in #325.

PR #321, opened 2026-05-10
Allow filter by "execution_type_desc" on Query Store screen

ClaudioESSilva:
## What does this PR do?

Implements #320

- [ ] Tests
- [ ] Documentation

## How was this tested?

1. Connect to Query Store
2. Choose `Execution Type` in the `Search by` combobox
3. Choose between `Regular`, `Aborted` or `Exception`
4. `Fetch` results

https://github.com/user-attachments/assets/02249aec-8d55-4bb2-9e96-6d173f75f432

## Checklist

### DIFF
--- a/src/PlanViewer.App/Controls/QueryStoreGridControl.axaml
+++ b/src/PlanViewer.App/Controls/QueryStoreGridControl.axaml
@@ -96,7 +96,7 @@
                              KeyDown="SearchValue_KeyDown"/>
                 </StackPanel>
                 <StackPanel x:Name="ExecutionTypePanel" Spacing="4" IsVisible="False">
-                    <TextBlock Text=" " FontSize="11"/>
+                    <TextBlock Text="Execution Type" Foreground="{DynamicResource ForegroundBrush}" FontSize="11"/>
                     <ComboBox x:Name="ExecutionTypeBox" Width="120" Height="36" FontSize="13"
                               SelectedIndex="0">
                         <ComboBoxItem Content="Regular" Tag="Regular"/>
--- a/src/PlanViewer.App/Controls/QueryStoreGridControl.axaml
+++ b/src/PlanViewer.App/Controls/QueryStoreGridControl.axaml
@@ -360,6 +360,7 @@
                         </DataTemplate>
                     </DataGridTemplateColumn.CellTemplate>
                 </DataGridTemplateColumn>
+                <DataGridTextColumn Header="Execution Type" Binding="{ReflectionBinding ExecutionTypeDesc}" SortMemberPath="ExecutionTypeDesc" Width="110"/>
             </DataGrid.Columns>
         </DataGrid>
         <!-- Loading overlay -->
--- a/src/PlanViewer.App/Controls/QueryStoreGridControl.axaml
+++ b/src/PlanViewer.App/Controls/QueryStoreGridControl.axaml
@@ -78,7 +78,7 @@
                 </StackPanel>
                 <StackPanel Spacing="4">
                     <TextBlock Text="Search by" Foreground="{DynamicResource ForegroundBrush}" FontSize="11"/>
-                    <ComboBox x:Name="SearchTypeBox" Width="120" Height="36" FontSize="13"
+                    <ComboBox x:Name="SearchTypeBox" Width="140" Height="36" FontSize="13"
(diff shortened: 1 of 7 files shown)


erikdarlingdata:
Thanks for this @ClaudioESSilva! A few things to address:

## Must-fix

- **`src/PlanViewer.Core/Services/QueryStoreService.cs` `FetchHistoryAsync`** — adds `MAX(rs.execution_type_desc)` to the SELECT (col 19) but the reader never reads it. The added SQL column is wasted work and `QueryStoreHistoryRow.ExecutionTypeDesc` will stay empty. (See next item — this method may be dead code.)

## Should-fix

- **Dead code edits.** `FetchHistoryAsync` and `FetchHistoryByHashAsync` appear to have no callers — `QueryStoreHistoryWindow` only uses `FetchAggregateHistoryAsync`. The PR edits all three but inconsistently (one path projects + reads, another projects but doesn't read). Could you either drop the changes to the dead methods or keep them consistent so we don't merge a half-wired version?

- **Grouped paths show empty Execution Type when no filter is set.** In `FetchGroupedByQueryHashAsync` and `FetchGroupedByModuleAsync`, the SQL doesn't project `execution_type_desc` from `#plan_stats`. Instead, after the query, the code does `foreach (var r in result.LeafRows) r.ExecutionTypeDesc = filter.ExecutionTypeDesc!`. When no filter is applied, every row in the grouped view shows blank. Recommended fix: project `MAX(rs.execution_type_desc)` from SQL like `FetchTopPlansAsync` does, for symmetry.

- **`AggregateGroupedRows` in `QueryStoreGridControl.axaml.cs`** doesn't carry `ExecutionTypeDesc` into the aggregated parent `QueryStorePlan`. Combined with the post-processing only writing to `LeafRows`/`IntermediateRows`, the parent (root) row in grouped views displays empty Execution Type even with a filter. Add `ExecutionTypeDesc = rows.FirstOrDefault()?.ExecutionTypeDesc ?? \"\"`.

- **`MAX(rs.execution_type_desc)` is alphabetically lossy.** \"Aborted\" < \"Exception\" < \"Regular\". A plan that ran both Regular and Aborted in the window will display \"Regular,\" masking the aborts — which is exactly the signal you'd want to see. Consider `STRING_AGG(DISTINCT rs.execution_type_desc, ', ')` or pick the value associated with `MAX(last_execution_time)` via a windowed function.

## Nits

- `ExecutionTypeBox` defaults to `SelectedIndex=0` (\"Regular\"). A user who selects \"Execution Type\" in the search-by combo and clicks Fetch without changing the sub-combo silently gets a Regular-only filter with no visible cue. Consider prepending an \"(any)\" item that maps to no filter.

- `ClearSearch_Click` resets `SearchTypeBox.SelectedIndex = 0` and `ExecutionTypeBox.SelectedIndex = 0` but doesn't directly reset panel visibility — it works only because the `SearchType_SelectionChanged` handler fires as a side effect. A one-line comment would make the dependency explicit.

- `Mcp/McpQueryStoreTools.cs` and `Cli/Commands/QueryStoreCommand.cs` don't expose an `execution_type_desc` filter, so MCP and CLI consumers can't take advantage of the new feature. Probably out of scope here, but worth a follow-up issue.

erikdarlingdata:
One more thought on the Search By options — could we add a **"Failed"** option that bundles `Aborted` + `Exception`? When troubleshooting, you usually want "what's not finishing successfully" rather than specifically aborts vs. exceptions. So the dropdown would be:

- Regular
- Aborted
- Exception
- Failed (= Aborted OR Exception)

In SQL that's `AND rs.execution_type_desc IN ('Aborted', 'Exception')` instead of the equality predicate. Easiest implementation: change `QueryStoreFilter.ExecutionTypeDesc` from `string?` to `string[]?` (or keep the string and split on a delimiter), and emit `IN` when there are multiple values.

ClaudioESSilva:
Thanks @erikdarlingdata !

I like that idea! Now we have all of the options 👍🏼 

My most common scenario is `aborted` with queries having an average duration of whatever the APP timeout is.

erikdarlingdata (approved):
All review feedback addressed cleanly. The lexicographic-concat trick to pick the most-recent execution_type_desc is a nice touch. LGTM!

Issue #322, opened 2026-05-10
[FEATURE] Improve Navigation via Keyboard/Mouse shortcuts


ClaudioESSilva:
### Which component(s) does this affect?

- [x] Desktop App
- [ ] CLI Tool
- [ ] SSMS Extension
- [ ] Plan Analysis Rules
- [ ] Documentation

### Problem Statement

Some common shortcuts that exist in other Windows tools are missing.

### Proposed Solution

Implement the following shortcuts:

## Using the keyboard
1. Allow navigate between open tabs using "CTRL + TAB" (forward) and "CTRL + SHIFT + TAB" (backwards).

## Using the mouse
1. On a mouse wheel-click on a tab, close it
2. On a double-click on a grid row that is grouped, expand that row (same as clicking on the arrow)

### Use Case

Using memory muscle! And it is convenient 😊 
I use a lot of shortcuts, and this helps me be quicker, too.

### Alternatives Considered

N/A

erikdarlingdata:
Implemented in #323 and merged to dev.

PR #323, opened 2026-05-10
feature - Improve navigation

ClaudioESSilva:
## What does this PR do?

Implements #322 

- [ ] Tests
- [ ] Documentation

### Tab navigation
1. Open multiple tabs with "CTRL + N"
2. Use "CTRL + TAB" and "CTRL + SHIFT + TAB" to navigate through the tabs.

### Close tabs with mouse wheel click
1. Open multiple tabs with "CTRL + N"
2. Use the mouse wheel to click on the tab text and see them closing.

### Expand/collapse grouped rows on Query Store grid
1. Open query store 
2. Double-click on a row to expand.
3. Double-click again to collapse.

## Checklist

### DIFF
--- a/src/PlanViewer.App/Controls/QueryStoreGridControl.axaml
+++ b/src/PlanViewer.App/Controls/QueryStoreGridControl.axaml
@@ -133,6 +133,7 @@
                   Background="{DynamicResource BackgroundDarkBrush}"
                   BorderThickness="0"
                   Sorting="ResultsGrid_Sorting"
+                  DoubleTapped="ResultsGrid_DoubleTapped"
                   ScrollViewer.HorizontalScrollBarVisibility="Auto">
             <DataGrid.ContextMenu>
                 <ContextMenu Opening="ContextMenu_Opening">
--- a/src/PlanViewer.App/Controls/QueryStoreGridControl.axaml.cs
+++ b/src/PlanViewer.App/Controls/QueryStoreGridControl.axaml.cs
@@ -7,6 +7,7 @@ using System.Runtime.CompilerServices;
 using System.Threading;
 using Avalonia.Controls;
 using Avalonia.Controls.Primitives;
+using Avalonia.Input;
 using Avalonia.Interactivity;
 using Avalonia.Media;
 using PlanViewer.Core.Interfaces;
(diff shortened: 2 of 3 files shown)


erikdarlingdata:
Nice — keyboard nav and middle-click close are great quality-of-life. One real bug and one nit.

## Should-fix

- **`Controls/QueryStoreGridControl.axaml.cs` `ResultsGrid_DoubleTapped`** — `DoubleTapped` on a DataGrid is a routed event that fires for taps anywhere inside the grid, including column headers, the empty area below the last row, and inner cell controls. Three problems:
  1. Double-clicking a **column header** (e.g. to auto-resize) toggles expansion of whichever row is currently selected.
  2. Double-clicking the **chevron `Button`** in the row produces a *triple* toggle: Click→Click on the button (toggles twice, ending in original state) plus the bubbled DoubleTapped on the grid (toggles once more) — net result is the opposite of what the user expects.
  3. Empty-area double-clicks below the last row also fire if a row is selected.

  Recommended fix: filter `e.Source` to require a `DataGridRow` ancestor and reject `Button` ancestors before toggling.

  ```csharp
  private void ResultsGrid_DoubleTapped(object? sender, TappedEventArgs e)
  {
      if (e.Source is not Visual v) return;
      if (v.FindAncestorOfType<Button>() != null) return;
      if (v.FindAncestorOfType<DataGridRow>() == null) return;
      if (ResultsGrid.SelectedItem is not QueryStoreRow row) return;
      if (!row.HasChildren) return;
      ToggleRowExpansion(row);
  }
  ```

## Nit

- **`MainWindow.axaml.cs` middle-click on tab header** — `StackPanel` with no background doesn't receive pointer events on the gap between its children. Hits register on `headerText` (TextBlock) and `closeBtn` (Button), and they bubble to the StackPanel handler, so it works in practice. To be defensive, set `Background=\"Transparent\"` on the header StackPanel so the whole rectangle is hit-testable.

## Clean

- Ctrl+Tab / Ctrl+Shift+Tab modifier matching is correct — the strict-equality `==` in the existing block doesn't trip because Ctrl+Shift+Tab is split into its own `else if` against `Control | Shift`. Wrap-around math handles 0/1 tabs and `SelectedIndex == -1` correctly.
- The `ExpandRow_Click` → `ToggleRowExpansion` extraction is a clean refactor.

erikdarlingdata (approved):
Both fixes look great — the DoubleTapped guards and the transparent header background are exactly the right shape. Thanks for the quick turnaround! LGTM.

PR #351, opened 2026-05-27
Feature/settings

rferraton:
## What does this PR do?

# Add unified Settings window

## Summary

Adds a non-modal Settings window (ALT+S) that centralises configuration previously scattered across hardcoded defaults. Settings persist to `%LocalAppData%\PerformanceStudio\appsettings.json`.

### New files

| File | Purpose |
|------|---------|
| `Dialogs/SettingsWindow.axaml` | AXAML shell — sidebar, button bar, ToggleSwitch styling |
| `Dialogs/SettingsWindow.axaml.cs` | Code-behind with three sections built programmatically |
| `Dialogs/FormatOptionRow.cs` | View-model for a single SQL format option row |

### Modified files

| File | Change |
|------|--------|
| `Services/AppSettingsService.cs` | Extended `AppSettings` with Query Store, Multi QS Overview, Query History, and Format Options properties. Added value clamping on load, `Debug.WriteLine` on deserialization failure, and legacy format-settings migration. |
| `Services/SqlFormatSettingsService.cs` | Removed the `SqlFormatSettingsService` façade (dead code after settings were unified). `SqlFormatSettings` class remains. |
| `MainWindow.axaml.cs` | Added `Settings_Click` handler, ALT+S key binding, singleton window management. |

### Deleted files

| File | Reason |
|------|--------|
| `Dialogs/FormatOptionsWindow.axaml` | Superseded by the Script Options section in SettingsWindow |
| `Dialogs/FormatOptionsWindow.axaml.cs` | Same — all functionality moved into SettingsWindow |

### Query Store
- Default history length (days), metric, top-N limit, time range, time display, group-by.

### Multi QS Overview
- Number of top databases (2–20) with per-slot hex colour picker and live preview.

### Query History
- Default chart metric, max plans fetched per query.

### Script Options (Format)
- Full `SqlFormatSettings` surface: keyword casing, SQL version, indentation, all newline/multiline/align flags.
- Boolean options rendered as toggle switches; choice options as combo boxes; numeric options as text fields.

## UI behaviour

- Non-modal singleton window — re-activates if already open.
- Sidebar `ListBox` switches between sections.
- **Save** writes to disk and raises `SettingsSaved` so the main window picks up changes immediately.
- **Cancel** prompts to discard if dirty.
- **Reset Section** resets only the visible section to defaults.
- **Reset All to Defaults** resets every setting (preserves app-state fields like recent/open plans).

## Bug fixes included

1. **Race condition** — Removed `SqlFormatSettingsService.Save` (read-modify-write on shared file) by deleting the dead façade.
2. **Visibility mismatch** — `FormatOptionRow` changed from `public` to `internal`; `PropertyInfo` was already `internal`.
3. **Swallowed exceptions** — `AppSettingsService.Load()` now logs failures via `Debug.WriteLine`.
4. **No input validation** — Numeric settings are clamped to valid ranges after deserialization.
5. **Dead code** — Removed `FormatOptionsWindow` (both files) and the unused `SqlFormatSettingsService` class.

## Testing

- Manual: open Settings, change values across all sections, save, relaunch, verify persistence.
- Manual: dirty-state prompt on cancel/close, Reset Section, Reset All.
- Build passes with no warnings.

- [ ] Tests
- [ ] Documentation

## How was this tested?

https://github.com/user-attachments/assets/61f544fb-8531-49f5-8d6b-c5ad37ee5159

https://github.com/user-attachments/assets/30b65e17-dbce-4a5c-9228-560926894149

- Plan files tested (estimated, actual, Query Store, etc.) : Estimated, actual, QS
- Platforms tested (Windows, macOS, Linux) : Windows only

## Checklist


erikdarlingdata (commented):
# Review — Unified Settings Window

Solid feature work overall — sensible architecture, good dirty-state handling, defaults are reasonable, value clamping on load is nice. A few real bugs and several smaller issues below.

---

## 🔴 Bugs

### 1. Reset Section mutates MainWindow's shared `_appSettings` even on Cancel
`MainWindow.Settings_Click` passes `_appSettings` **by reference** into `new SettingsWindow(_appSettings)`. `Reset_Click` (sections 0 and 1) then writes directly to `_settings.QueryStoreSlicerDays = …`, etc. — which mutates that same shared instance. If the user clicks **Reset Section** then **Cancel**, the in-memory `MainWindow._appSettings` is left with reset values that were never persisted to disk. `ResetAll_Click` is safe because it rebinds `_settings = fresh;`, but section reset isn't.

**Fix:** stage edits to a local working copy and only assign to `_settings` (and invoke `SettingsSaved`) inside `Save_Click`. Or pass a deep clone into the SettingsWindow ctor.

### 2. `QuerySessionControl.Format.cs` — broken indentation after the patch
The diff replaced the trailing braces with:
```csharp
        finally
        {
                    FormatButton.IsEnabled = true;
                    }
                }
            }
```
The brace **count** is correct (closes `finally`/method/class), but indentation is 20/20/16/12 spaces — clearly a botched edit. Compiles, but ugly and inconsistent with the rest of the file. Re-indent to 12/8/4.

### 3. `SelectComboByTag` silently leaves `SelectedIndex = -1` on unknown tag
In `QueryStoreGridControl`, if `userSettings.QueryStoreDefaultMetric` is a stale or corrupted value, `OrderByBox` ends up with no selection. The `GroupByBox` call has a `_ => \"query-hash\"` fallback, but `OrderByBox` does not. Add a fallback or assert a default index.

### 4. Legacy format-settings file is orphaned when `FormatOptions` already exists
`MigrateFormatSettings` early-returns if `settings.FormatOptions != null`, leaving the old `perfstudio_format_settings.json` on disk forever. Minor, but worth deleting unconditionally when the file is present and either migrated or superseded.

---

## 🟡 Concerns

### 5. `AppSettingsService.Load()` is called many times per Query Store view
New call sites: `QueryStoreGridControl` ctor, `QueryStoreOverviewControl` ctor, `QueryStoreHistoryControl.BuildColorMap` (called on every history refresh), `TimeRangeSlicerControl` field initializer, `QueryStoreHistoryWindow` ctor. Each call hits disk + deserializes JSON. Not catastrophic, but `BuildColorMap` runs per refresh, which is wasteful. Consider caching `AppSettings` in-process (with invalidation via `SettingsSaved`).

### 6. `TimeRangeSlicerControl._activeFilterTag` initialized via field initializer
```csharp
private string _activeFilterTag = AppSettingsService.Load().QueryStoreDefaultTimeRange;
```
Field initializers run before the constructor body and before `InitializeComponent` — lower risk for a static disk read, but moving into the constructor is more conventional.

### 7. `Reset Section` for the Script Options section discards `FormatOptions` to `null` then rebuilds rows
Works because `BuildScriptOptionsSection` uses `_settings.FormatOptions ?? new SqlFormatSettings()`. Fine — but `_isDirty = true` is set inside Reset before the user explicitly intends to save, so a Cancel after that leaks intent. (Same bug class as #1.)

### 8. Hex color input has no validation on save
`_colorTextBoxes.Select(tb => tb.Text ?? \"#555555\")` persists whatever the user typed. Garbage values silently fall back to gray at draw time via the `try { Color.Parse } catch` swallow. At least show a red border or block save when any value fails to parse — currently the user gets no feedback.

### 9. `AppSettingsService.MigrateFormatSettings` calls `Save(settings)` inside `Load()`
A read operation that may write to disk is surprising. If `Save` throws inside `MigrateFormatSettings`, the catch swallows it and the old file stays, so the migration retries on every Load. Reasonable in practice but worth a comment.

### 10. Doc comment for `AccuracyRatioDivergenceLimit` was dropped in the move
The XML summary above it was removed when the field was relocated to the \"App State\" block. Re-attach.

---

## 🟢 Style nits

- `SettingsWindow.axaml.cs` uses tabs; the rest of the project appears to use 4 spaces. Confirm against `.editorconfig`.
- `SettingsWindow.axaml` ResetAll/Reset buttons have an extra level of indentation in the source vs. the Save/Cancel buttons (visual only).
- `_colorTextBoxes`/`_colorPreviews` are rebuilt on every `_topDbCountBox` change — fine, but previous handlers on old TextBoxes are unreferenced and GC'd, not actively detached.

---

## ✅ Things done well

- Sensible value clamping on load.
- `Debug.WriteLine` instead of swallowing — much better than the prior behavior.
- Singleton pattern in `Settings_Click` is correct (UI-thread only, no race).
- Migration logic is conservative (only fires when `FormatOptions == null` and old file exists).
- Default color palette extracted as a single source of truth (`DefaultTopDbColors`).
- Removing `SqlFormatSettingsService.Save` eliminates the read-modify-write race against the shared file.

---

The bugs in #1 and #2 are worth addressing before merge; the rest can land as follow-ups.

erikdarlingdata:
Re-reviewed commit 375cf3b — looks great. All four 🔴 bugs and three of the four 🟡 concerns are fixed:

- ✅ #1 Reset leaks → `SettingsWindow` ctor now deep-clones via JSON roundtrip
- ✅ #2 indentation
- ✅ #3 `SelectComboByTag` fallback to index 0
- ✅ #4 legacy file deleted unconditionally
- ✅ #5 `_cached` + `Invalidate()` (heads up: `Load()` returns the shared cached instance, so callers must not mutate it — worth an XML doc note when you next touch the file)
- ✅ #6 ctor instead of field initializer
- ✅ #7 fixed as a side effect of #1
- ✅ #8 red border + short-circuited save on invalid hex
- ✅ #9 explanatory comment on Save-inside-Load
- Only nit left is #10 (missing XML doc on `AccuracyRatioDivergenceLimit`) — not a blocker.

Nice work, ready from my side.

Issue #139, opened 2026-03-24
[BUG] Application Does not open on Linux


steveearle86:
### Component

Desktop App (Linux)

### SQL Performance Studio Version

1.2.5

### Operating System

Linux Mint 22.3, Ubuntu 24.04.4 LTS, Debian 13 (stable)

### Describe the Bug

Upon opening the application, I get an unhandled exception (below) the issue appears to be that an older version of the SkiaSharp library (88.1) is included in the artifacts but the application requires 119.0 

> Unhandled exception. System.TypeInitializationException: The type initializer for 'SkiaSharp.SKFontManager' threw an exception.
 ---> System.TypeInitializationException: The type initializer for 'SkiaSharp.SKObject' threw an exception.
 ---> System.InvalidOperationException: The version of the native libSkiaSharp library (88.1) is incompatible with this version of SkiaSharp. Supported versions of the native libSkiaSharp library are in the range [119.0, 120.0).
   at SkiaSharp.SkiaSharpVersion.CheckNativeLibraryCompatible(Version minSupported, Version current, Boolean throwIfIncompatible)
   at SkiaSharp.SkiaSharpVersion.CheckNativeLibraryCompatible(Boolean throwIfIncompatible)
   at SkiaSharp.SKObject..cctor()
   --- End of inner exception stack trace ---
   at SkiaSharp.SKObject..ctor(IntPtr handle, Boolean owns)
   at SkiaSharp.SKFontManager..ctor(IntPtr handle, Boolean owns)
   at SkiaSharp.SKFontManager.SKFontManagerStatic..ctor(IntPtr x)
   at SkiaSharp.SKFontManager..cctor()
   --- End of inner exception stack trace ---
   at SkiaSharp.SKFontManager.get_Default()
   at Avalonia.Skia.FontManagerImpl..ctor()
   at Avalonia.Skia.SkiaPlatform.Initialize(SkiaOptions options)
   at Avalonia.SkiaApplicationExtensions.<>c.<UseSkia>b__0_0()
   at Avalonia.AppBuilder.SetupUnsafe()
   at Avalonia.AppBuilder.Setup()
   at Avalonia.AppBuilder.SetupWithLifetime(IApplicationLifetime lifetime)
   at Avalonia.ClassicDesktopStyleApplicationLifetimeExtensions.StartWithClassicDesktopLifetime(AppBuilder builder, String[] args, Action`1 lifetimeBuilder)
[...]

Issue #391, opened 2026-07-17
[FEATURE] Query Store server-side filter panel, chips, and sp_QuickieStore-parity pre-filters


erikdarlingdata:
Follow-up to #389 (query-text search). Adds the full server-side filtering architecture discussed there:

- Stackable pre-TOP-N filter criteria in all three Query Store SQL builders: min executions, min avg duration/CPU (ms), query-text exclude, escape-brackets toggle, include/ignore query-id and plan-id lists
- Desktop: collapsible Server Filters panel, active-filter chip strip (removable, re-fetches), and promotion of column filters to server searches
- MCP get_query_store_top and CLI query-store parity for all new criteria

Generated with [Claude Code](https://claude.com/claude-code)

PR #395, opened 2026-07-25
Quarterly maintenance: CVE-2025-6965 fix, dependency bumps, zero-warning build, CI action bumps

erikdarlingdata:
Quarterly maintenance pass (July 2026). Read-only scans + low-risk fixes per the /maintenance skill.

## Security
- **CVE-2025-6965 (High, GHSA-2m69-gcr7-jv3q)**: PlanShare pulled SQLitePCLRaw.lib.e_sqlite3 2.1.11 transitively via Microsoft.Data.Sqlite (SQLite < 3.50.2 memory corruption). Fixed with a direct `SQLitePCLRaw.bundle_e_sqlite3` 2.1.12 reference (Microsoft.Data.Sqlite 10.0.10 still floors at the vulnerable 2.1.11). NU1903 no longer fires.

## Dependency bumps (patch/minor only)
| Package | From | To |
|---|---|---|
| Avalonia (App, 5 pkgs) | 11.3.17 | 11.3.18 |
| Microsoft.Data.SqlClient | 7.0.1 | 7.0.2 |
| Microsoft.Data.SqlClient.Extensions.Azure | 1.0.0 | 7.0.2 (version-scheme alignment w/ SqlClient) |
| Microsoft.SqlServer.TransactSql.ScriptDom | 180.37.3 | 180.59.2 |
| ModelContextProtocol (+AspNetCore) | 1.4.0 | 1.4.1 |
| System.CommandLine | 2.0.9 | 2.0.10 |
| AspNetCore WASM (+DevServer) | 10.0.9 | 10.0.10 |
| Microsoft.NET.Test.Sdk | 18.6.0 | 18.8.1 |
| Microsoft.Data.Sqlite (PlanShare) | 10.0.9 | 10.0.10 |

**Held back on purpose:** ScottPlot.Avalonia stays 5.1.58 — 5.1.59+ requires Avalonia >= 12 (NU1605s against 11.x). That means the parked `upgrade/avalonia-12` branch is now unblocked; it gets its own effort. `Microsoft.Win32.Registry` removed (in-box on net10.0, NU1510). SkiaSharp.NativeAssets.Linux pin (issue #139 guard) unchanged. VSIX stays on the 17.x tip (17.14.x is the line's latest; 18.x targets VS 18).

## Zero-warning build
Avalonia 11.3.18 obsoletes `DragEventArgs.Data` / `DataFormats.Files` (CS0618 x6) — migrated `MainWindow.FileOps.cs` drag/drop to `e.DataTransfer.Contains(DataFormat.File)` / `TryGetFiles()` (API verified against Avalonia 11.3.18 source).

## CI actions
checkout v5->v7, setup-dotnet v5->v6, upload-artifact v6->v7, upload-pages-artifact v3->v5, deploy-pages v4->v5. setup-msbuild@v3 and signpath@v2 already current. No node-deprecation annotations were pending; vpk pin (1.2.0) still matches the Velopack PackageReference.

## Test plan
- [x] Clean Debug build of PlanViewer.sln + PlanShare: **0 warnings, 0 errors**
- [x] 182/182 Core tests pass
- [x] Publish smoke linux-x64 + osx-arm64: apphost + correct SkiaSharp natives present
- [x] `dotnet list package --vulnerable --include-transitive`: PlanShare NU1903 cleared, solution clean
- [ ] CI (this PR) validates checkout@v7 + setup-dotnet@v6; nightly + pages deploy validate the rest on their next runs

### DIFF
--- a/.github/workflows/check-pr-branch.yml
+++ b/.github/workflows/check-pr-branch.yml
@@ -6,6 +6,10 @@ on:
       - reopened
       - synchronize
       - edited
+
+# pull_request_target runs with the base repo's token — this job needs none of it.
+permissions: {}
+
 jobs:
   check-branches:
     runs-on: ubuntu-latest
--- a/.github/workflows/check-version-bump.yml
+++ b/.github/workflows/check-version-bump.yml
@@ -10,7 +10,7 @@ jobs:
 
     steps:
       - name: Checkout PR branch
-        uses: actions/checkout@v5
+        uses: actions/checkout@v7
 
       - name: Get PR version
         id: pr
@@ -26,7 +26,7 @@ jobs:
           Write-Host "PR version: $version (from $path)"
 
       - name: Checkout main
-        uses: actions/checkout@v5
+        uses: actions/checkout@v7
         with:
           ref: main
           path: main-branch
--- a/.github/workflows/check-version-bump.yml
+++ b/.github/workflows/check-version-bump.yml
@@ -3,6 +3,9 @@ on:
   pull_request:
     branches: [main]
 
+permissions:
+  contents: read
+
 jobs:
   check-version:
     if: github.head_ref == 'dev'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,9 @@ on:
       - 'src/PlanViewer.Ssms/**'
       - 'src/PlanViewer.Ssms.Installer/**'
 
+permissions:
+  contents: read
+
 jobs:
   build-and-test:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,10 +24,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v7
 
       - name: Setup .NET 10.0
-        uses: actions/setup-dotnet@v5
+        uses: actions/setup-dotnet@v6
         with:
           dotnet-version: 10.0.x
           cache: true
--- a/.github/workflows/deploy-web.yml
+++ b/.github/workflows/deploy-web.yml
@@ -27,10 +27,10 @@ jobs:
       url: ${{ steps.deployment.outputs.page_url }}
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v7
 
       - name: Setup .NET 10.0
-        uses: actions/setup-dotnet@v5
+        uses: actions/setup-dotnet@v6
         with:
           dotnet-version: 10.0.x
 
@@ -50,10 +50,10 @@ jobs:
         run: cp publish/wwwroot/index.html publish/wwwroot/404.html
 
       - name: Upload Pages artifact
-        uses: actions/upload-pages-artifact@v3
+        uses: actions/upload-pages-artifact@v5
         with:
           path: publish/wwwroot
 
       - name: Deploy to GitHub Pages
         id: deployment
-        uses: actions/deploy-pages@v4
+        uses: actions/deploy-pages@v5
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -15,7 +15,7 @@ jobs:
     outputs:
       has_changes: ${{ steps.check.outputs.has_changes }}
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v7
         with:
           ref: dev
           fetch-depth: 0
@@ -38,12 +38,12 @@ jobs:
     runs-on: windows-latest
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v7
         with:
           ref: dev
 
       - name: Setup .NET 10.0
-        uses: actions/setup-dotnet@v5
+        uses: actions/setup-dotnet@v6
         with:
           dotnet-version: 10.0.x
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: windows-latest
 
     steps:
-      - uses: actions/checkout@v5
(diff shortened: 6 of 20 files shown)


erikdarlingdata:
---

## Update: security review fixes (second commit)

The pass also ran a `security-review` on `git diff f57fad8..HEAD` (120 commits). **10 findings: 1 High, 2 Medium, 7 Low.** Severities are calibrated per the maintenance skill (single-user laptop; loopback/local-IPC is Low here). Fixed the High, both Mediums, and two mechanical Lows in `04ae735`.

**The Query Store server-side filter panel (#391/#392) — the prime suspect — is clean.** Every user value becomes a `SqlParameter` (including id lists, expanded to `@includeQueryId0, @includeQueryId1, ...`); the only interpolated fragments are a whitelisted metric column, integers, and a bool-derived `ASC`/`DESC`.

### HIGH — MCP DNS rebinding (fixed)
`Mcp/McpHostService.cs` — Kestrel binds loopback, but nothing validated `Host`/`Origin`, so a malicious page could rebind a DNS name it controls to 127.0.0.1 and talk to the MCP server **same-origin** (CORS never applies), then call `get_connections` / `get_query_store_top` / `get_plan_xml` to pull saved server names, query text, and plan XML. Precondition is MCP enabled (off by default) — but once on, no further user action. This is the remotely-reachable carve-out the calibration reserves High for, not a local-IPC finding.

Fixed with middleware ahead of `MapMcp()` that 403s non-loopback `Host` and non-loopback `Origin`. **Runtime-verified against the running app:**

| Request | Result |
|---|---|
| `Host: evil.example.com` | **403** |
| `Origin: http://evil.example.com` | **403** |
| Legit loopback `initialize` | **200** + full server handshake |

### MEDIUM — plan-derived parameters spliced into an executed batch (fixed)
`ReproScriptBuilder.cs` — parameter names/types/values come straight off untrusted plan XML and landed raw in the generated `EXECUTE sys.sp_executesql`, which the app executes (`ActualPlanExecutor`) and `get_repro_script` serves. A crafted `.sqlplan` with `ParameterCompiledValue="1; <T-SQL> --"` escaped the very parameterization the code builds. Names/types must now match a plain identifier/type pattern (else the parameter is dropped) and values must be a single self-contained literal (else the existing `?` placeholder). Pre-existing, not introduced this window.

### MEDIUM — identifier escaping in scripted DDL (fixed)
`DdlScripter.cs` / `SchemaQueryService.cs` — `BracketName` never doubled embedded `]` and passed through anything already starting with `[`; index key/include column lists weren't bracketed at all. An object named `x] ; <T-SQL> --` produced a script that runs attacker SQL when copied and executed, which is that output's whole purpose. Now doubles `]` (mirroring `ReproScriptBuilder`) and `QUOTENAME`s the `FOR XML PATH` column lists.

> Note: `FormatIndexes_AlreadyBracketedName_IsNotDoubleBracketed` pinned the passthrough that *was* the bypass, so it now asserts correct escaping instead. Catalog names are always raw, so a leading `[` is part of the name, not pre-quoting.

### LOW — fixed (mechanical)
- `permissions: {}` on the `pull_request_target` workflow (was uncapped), `contents: read` on ci/check-version-bump.
- `Uri.EscapeDataString` on share id + delete token in `PlanShareService`.

### LOW — reported, not fixed (need your call)
1. **"Get Actual Plan" executes a loaded plan file's embedded SQL** after only a connection prompt; the query-session path *does* confirm first (`MainWindow.PlanViewer.cs:375-478` vs `QuerySessionControl.Execution.cs:251-255`). Adding a dialog is a UX decision, and it's identical at baseline — not a regression.
2. **PlanShare**: plan GET has no rate limiter (only `/api/share` and `/api/event` do) and `/api/stats` has no auth — exposes share counts, traffic, referrers. The 8-char id (~2^47) keeps this resource-abuse, not a confidentiality break. Depends on whether the stats dashboard is meant to be public.
3. **PlanShare `visitor_hash`** is `SHA256(ip|ua|day)` truncated to 16 hex — brute-forceable back to the source IP (IPv4 is 2^32), which contradicts the adjacent "no PII stored" comment. Fix is a server-side secret salt; touches the deployed server, so it's not a same-PR change.
4. **SSMS `AppLauncher`** tries `PATH` before well-known install locations (a writable `PATH` entry wins), and leaves `%TEMP%\ssms_plan_*.sqlplan` files containing query text behind. The VSIX isn't in `PlanViewer.sln` or CI and I can't smoke-test it here.

### Verified clean
`XDocument.Parse` prohibits DTDs by default on .NET 10 (no XXE) plus an explicit depth cap; schema lookups parameterize via `OBJECT_ID(@objectName)`; no hardcoded secrets; credentials in Credential Manager/Keychain, not `settings.json`; `TrustServerCertificate` false + `Encrypt` Mandatory; `HtmlExporter` encodes interpolated plan content; share id/token generation correctly uses `RandomNumberGenerator`; `zip_with_exec.py` has no zip-slip.

**Tests: 190 pass** (8 new covering both injection paths), clean Debug build still 0 warnings.

erikdarlingdata:
---

## Update: removed the last deprecated package (`26833b2`)

`Avalonia.Diagnostics` was the solution's only deprecated package (NuGet "Legacy", superseded by `AvaloniaUI.DiagnosticsSupport`) — and it turned out to be **completely unused**: there is no `AttachDevTools()` call anywhere in the tree, so the DevTools inspector it provides was never reachable. Upstream's own notes say the old package "can be safely removed, as it is not used by new Developer Tools", and the parked Avalonia 12 branch had already dropped it for the same reason. Removed rather than swapped.

`dotnet list package --deprecated` is now **empty across every project**.

Also audited the `NoWarn` list per the maintenance checklist: `AVLN3001` is the only suppression, and clearing it still produces 4 warnings (QuerySessionControl, QueryStoreGridControl, QueryStoreOverviewControl, ConnectionDialog), so its documented reason is still accurate. **Kept.**

**Re-validated after the removal:** Debug *and* Release both build with 0 warnings, 190/190 tests pass, and the app launches and paints its window (3s).

PR #411, opened 2026-07-25
Never publish a release before signing; harden PlanShare

erikdarlingdata:
Two things: stop `release.yml` publishing before signing, and the three remaining PlanShare items from the maintenance security review.

## release.yml: nothing is published until signing succeeds

`gh release create` ran as step 5, before signing at step 15. So when signing failed, the release was **already public** - with no app assets on it. That is exactly what happened to v1.17.0 when the SignPath approval window lapsed: a published release carrying only the VSIX.

Since SignPath offers **manual approval only**, this was guaranteed to recur.

The release is now created after signing succeeds. The two SSMS uploads moved with it, because uploading requires the release to exist - that dependency is the reason the original ordering existed, and it is handled by moving both rather than by using a draft (a draft has no git tag until published, and `vpk upload github` resolves the release by tag).

New order:

```
 8. Build SSMS extension        (built early, not uploaded)
11. Sign Windows build          <- everything above is reversible
13. Create release              <- first thing that publishes
14. Upload SSMS extension
15. Publish to SSMS Gallery
16. Create Velopack release
17. Package and upload
```

A failed or unapproved signing now leaves **no trace** - no tag, no release, nothing for users to stumble into.

## PlanShare hardening

**Rate limit the plan GET** (120/min per IP). It was the only unlimited endpoint, so nothing bounded scraping or id enumeration. 120/min is far above any human browsing rate.

**Salt `visitor_hash`.** It was `SHA256(ip|ua|day)` truncated to 16 hex - brute-forceable straight back to the source IP, since IPv4 is only 2^32, the User-Agent is guessable and the day is known. The adjacent "no PII stored" comment was therefore not true. Now `HMAC-SHA256` with a 32-byte secret generated once into `data/visitor-salt.bin` - kept **beside** the database rather than inside it, so a copy of `plans.db` does not carry the means to reverse its own hashes.

**Optional auth on `/api/stats`**, which exposes share counts, daily traffic, unique visitors and top referrers. Deliberately opt-in: enabled only when `STATS_TOKEN` is set, so deploying this cannot break the dashboard. Accepts the token by `X-Stats-Token` header or `?token=`, compared with `CryptographicOperations.FixedTimeEquals` so it cannot be recovered a byte at a time.

## Verified against a running instance

| Check | Result |
|---|---|
| `/api/stats` no token | **401** |
| `/api/stats` wrong token | **401** |
| `/api/stats` correct token (header) | **200** |
| `/api/stats` correct token (query) | **200** |
| share + read still work | **200** |
| GET rate limiter, 130 requests | **119 x 200, 11 x 429** |
| salt file on first start | created, 32 bytes |

Plus a clean Release build with 0 warnings, and `release.yml` step ordering asserted programmatically.

## Deployment

`server/PlanShare/**` auto-deploys on push to `main` (#399), so these ship with the next release. `STATS_TOKEN` is not set on the server, so `/api/stats` stays public until you choose to set it - deliberate, so this deploy cannot break anything.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,13 +37,11 @@ jobs:
             echo "EXISTS=false" >> $GITHUB_OUTPUT
           fi
 
-      - name: Create release
-        if: steps.check.outputs.EXISTS == 'false'
-        shell: bash
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          gh release create "v${{ steps.version.outputs.VERSION }}" --title "v${{ steps.version.outputs.VERSION }}" --generate-notes --target main
+      # NOTE: the release is deliberately NOT created here. It is created after
+      # signing succeeds, further down. Creating it up front meant a signing
+      # failure left a published, publicly-visible release with no app assets -
+      # which is exactly what happened to v1.17.0 when the SignPath approval
+      # window lapsed. Nothing before the signing step publishes anything now.
 
       - name: Setup .NET 10.0
         uses: actions/setup-dotnet@v6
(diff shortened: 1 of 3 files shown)


erikdarlingdata:
Both review findings addressed, plus a correction to my own PR description.

**1. `readRateLimiter` never swept** — real bug, mine, introduced in this PR. It was not in the `RateLimiters` record so `CleanupService` never called `Sweep()` on it, leaving one permanent dictionary entry per unique client IP for the process lifetime. Registered and swept alongside the other two.

**2. Rate limiters were a global cap, not per-IP** — pre-existing, but my read limiter turned it from latent into an availability risk. Confirmed on the server: nginx *does* set `X-Real-IP` and `X-Forwarded-For`, but the app never read them, so every `RemoteIpAddress` was `127.0.0.1` and all three limiters shared one bucket. One busy client would have throttled everyone.

Added `UseForwardedHeaders` with `KnownProxies` restricted to loopback — trusting `X-Forwarded-For` from arbitrary sources would let any caller spoof its address and bypass the limits entirely, which is worse than the original bug.

Verified per-IP behaviour:

| Client | Result |
|---|---|
| A (`203.0.113.10`), 125 requests | 120 x 200, then 5 x 429 |
| B (`198.51.100.77`), first request | **200** — unaffected by A's exhausted quota |

Before the fix, B would have been rejected.

**3. Dashboard could not authenticate** — also correct, and the failure mode was the bad part: `loadPlanStats()` swallowed non-200s and rendered "No data yet", identical to an empty database. Enabling the very feature this PR adds would have silently blanked the dashboard.

The fetch now sends `X-Stats-Token`, bootstrapped once from `?token=...`, stored in `localStorage`, and stripped from the URL with `replaceState` so it does not linger in the address bar or history. A 401 renders a distinct message naming the cause and the fix; other non-2xx report their status. Both via `textContent`, not `innerHTML` — this dashboard already had an XSS fix (#249) and building HTML from response-derived strings invites it back.

Verified with `STATS_TOKEN` set: no header 401, header 200, `?token=` 200. Inline script extracted and `node --check`ed.

**Correction to my description:** I wrote that the release.yml step ordering was "asserted programmatically". That overstates it — it was a one-off script I ran locally, not persisted as a test. There is no regression coverage for that ordering, and the reviewer was right to question it. Treat the ordering as verified once, by inspection, not guarded.

Also switched `KnownNetworks` to `KnownIPNetworks`; the former is obsolete in .NET 10 and emitted `ASPDEPR005`, and the standard here is 0 warnings.

Issue #415, opened 2026-07-28
[BUG] Robot Advice Copy to Clipboard can crash app when clipboard is unavailable


SalmanRajwani:
### Component

Desktop App (Windows)

### SQL Performance Studio Version

1.19.0

### Operating System

Windows 11

### Describe the Bug

Clicking **Copy to Clipboard** in the Robot Advice window can crash the app when the Windows clipboard is temporarily unavailable/busy.

This appears to be an unhandled async clipboard exception in the Robot Advice copy handler. The plan-analysis pipeline itself completed successfully locally, and the Robot Advice window/content also built and rendered successfully. The failure evidence points specifically to the clipboard write path.

### Steps to Reproduce

1. Open a valid execution plan.
2. Open **Robot Advice**.
3. Click **Copy to Clipboard** while the Windows clipboard is unavailable or busy.

### Expected Behavior

The app should not terminate. Clipboard failures should be caught and surfaced as a non-fatal UI message, for example: "Clipboard is busy, try again."

### Actual Behavior

The app can terminate with a Windows `APPCRASH`.

Observed Windows crash details:

```text
Faulting application name: PlanViewer.App.exe
Faulting application version: 1.19.0.0
Faulting module name: KERNELBASE.dll
Exception code: 0xe0434352
```

Managed exception stack captured locally:

```text
System.Runtime.InteropServices.COMException: OpenClipboard Failed (0x800401D0 (CLIPBRD_E_CANT_OPEN))
   at Avalonia.Win32.ClipboardImpl.SetDataAsync(IAsyncDataTransfer dataTransfer)
   at PlanViewer.App.Services.AdviceWindowHelper.<>c__DisplayClass0_0.<<Show>b__2>d.MoveNext()
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__124_0(Object state)
   at Avalonia.Threading.SendOrPostCallbackDispatcherOperation.InvokeCore()
```

### Plan File

Not attached. The issue does not appear to be plan-specific, and I do not want to share query text or schema details publicly.

### Additional Context

I checked for existing issues mentioning `OpenClipboard`, `CLIPBRD_E_CANT_OPEN`, and Robot Advice clipboard crashes and did not find a duplicate. The closest historical issue I found was #81, but that was for Robot/Human Advice windows not opening and Copy Repro Script behavior in 1.1.0, and it is closed.

erikdarlingdata:
[...]

PR #518, opened 2026-09-14
Phase 1 UI/UX fixes from the live review

erikdarlingdata:
Phase 1 of the UI/UX modernization pass: the bugs and quick wins found by a live click-through review of v1.25.0 (driven session against real plans and SQL2025/StackOverflow2013), built by four parallel workstreams and integrated here.

## Bugs fixed

- **Settings always claimed unsaved changes** — opening Settings and clicking Cancel with zero interaction prompted "Discard changes?" every time. Dirty-tracking handlers fired during programmatic control initialization; now guarded through construction and the layout pass (with a build token, since section 0 builds twice on open). Also: the Settings window is now a real modal (`ShowDialog`), closing the leak where the main window's menus could open on top of its discard prompt; the close walk is latched against re-entrancy; the discard dialog completes on `Closed`, not `Closing`.
- **Raw exception text parked in the toolbar forever** — "A task was canceled." from a view switch sat in the status strip indefinitely. Cancellations are now swallowed entirely; real errors render in a new `ErrorBrush` token and auto-clear after 12s; the strip clears on sub-tab change and on detach (the second permanent-park path).
- **Compare Plans silently no-oped** with fewer than two plans open. Both Compare buttons (session toolbar and plan-tab toolbar) now disable with an explanatory tooltip and re-evaluate as tabs open/close, including detached plan windows.
- **Column filter popup showed internal property names** ("Filter: QueryHash" → "Filter: Query Hash").
- **Connect dialog race** (found during the work): editing the server field while a connect was in flight could save and return a server that was never validated. Connect now captures its inputs before awaiting and re-checks the dialog is still open after.

## UX fixes

- **Empty query tabs now offer the primary actions**: Open a plan (Ctrl+O), Paste plan XML (Ctrl+V), Connect to a server, and the 5 most recent plans. Hidden as soon as the editor has text or a plan is open; shortcuts shown are pinned by test to the real File-menu gestures.
- **Connecting is one step**: Connect is enabled from field state and performs connect + database enumeration itself; Test Connection remains as an optional check. Reconnect preselects the session's current database.
- **Double-click loads a plan from the Query Store grid** (same path as the context menu's Load Plan); group rows keep expand/collapse.
- **Embedded plan viewers no longer duplicate the session toolbar** — the dead Reconnect/server/Database row is gone from plans opened inside a query session; standalone plan tabs keep theirs.
- **The "Query Editor" sub-tab no longer renders at 24px** next to its 12px siblings.

## Properties panel overhaul

The panel Erik called out as the most glaringly clunky part of the UI:

- Opens at 380px with a visible, hover-highlighted splitter; dragged width is remembered instead of resetting to 320px on every node click.
- Per-thread stats no longer flatten into the list (a DOP-8 plan produced hundreds of "Thread N" rows). Each actual section gets one collapsed "Per-thread breakdown" expander, with row-skew flagged in the header.
- Predicates, residuals, and output lists render full-width in monospace under their label instead of wrapping into a ~180px bracket tower.
- A filter box under the header narrows 60+ rows by label or value, sticky across node selections.
- A real copy menu: Copy value / Copy name and value / Copy all properties — replacing the default text-box menu that offered a disabled Copy and an enabled Paste on read-only data.

## Number formatting

New shared `MetricFormatter` in Core (36 unit tests): costs render at ≤4 significant decimals with trailing zeros stripped ("3526.210000" → "3526.21", "0.000000" → "0", tiny-but-real costs as "<0.0001" so they never read as free); durations auto-scale on the existing ms/s/m ladder. Applied to the properties panel, tooltips, statements grid, comparison report, and the Web viewer (linked source, so both viewers agree). Query Store grid blanks Query ID / Plan ID on grouped rows instead of showing 0, and the copy-ID menu items disable accordingly. Machine-readable outputs (Robot Advice JSON, MCP, plan XML) untouched at full precision.

## Test plan

- [ ] Release build: zero warnings; `PlanViewer.Core.Tests` green (574 expected)
- [ ] Settings: open → Cancel (no prompt); change a value → Cancel (prompts); Script Options → scroll to bottom → Cancel (no prompt)
- [ ] Live re-drive of every fixed flow with the screenshot harness (adversarial pass, results to follow as a PR comment)

https://claude.ai/code/session_017DEJ8FiSKJg44Ew2GG2nQ7


erikdarlingdata:
## Adversarial review, round 1 (code)

Two passes ran against the integrated branch: a full-diff review and an Avalonia-trap review focused on the cross-workstream seams (four authors touched overlapping files). Results:

**1 Medium, found and fixed (`ea10645`)** — two individually-correct changes composed into silent error loss: plan-load failures became auto-clearing error statuses (B2 work) while sub-tab selection began clearing the status strip (also B2 work), so a bad plan in the middle of a Query Store multi-select batch had its error wiped by the next plan's tab selection, and the summary was suppressed because not every plan loaded. The batch now writes one authoritative status after the loop; pinned by three new headless tests (mixed batch, clean batch, single failure).

**3 Lows, fixed (`3e05501`)** — a status-clear timer race that could wipe a fresh message; the QS results grid starting a pointless cell edit on the second click of the new double-click-to-load gesture (grid is now read-only; template chevron/checkbox unaffected); empty-state rows firing on right/middle click.

**1 app-wide gap, fixed (`91a2ec5`)** — the properties workstream discovered `SelectableTextBlock` registers its own copy event, outside the #415 clipboard guard, and fixed it inside its own file fence; the app-wide class handler now covers the remaining unguarded instances (advice views, execution results, missing-index panels).

**Explicitly verified clean:** the two-author merge of PlanViewerControl (hosted-mode vs properties rewrite, single writer for the panel column, min/max clamp cleared before close); the settings dirty-guard (all format rows realize inside the guard window — no scroll-time false dirty); the connect dialog's capture-before-await; ErrorBrush theming via DynamicResource; SelectableTextBlock hit-test/flyout/Ctrl+C traps in the properties panel.

Gate after fixes: solution build 0 warnings, 577 tests / 0 failed / 1 pre-existing skip.

**Still owed: the live re-drive** — every fixed flow driven in the running app (screenshot harness), including the visual judgments no headless test can make (380px default panel width, splitter grab feel, nested breakdown expander, error-red rendering). Results will follow as a second comment.

https://claude.ai/code/session_017DEJ8FiSKJg44Ew2GG2nQ7

erikdarlingdata:
## Adversarial review, round 2 (live re-drive)

Every fixed flow driven in the running app at 3840x2400, scripted input + screenshots. **All pass.**

| Finding | Live result |
|---|---|
| B1 settings false-dirty | Open→Cancel silent; real change→Cancel prompts; Script Options→scroll to bottom→Cancel silent (the flagged guard edge holds) |
| B4 modality | Help menu refused to open over the discard prompt |
| B2 status handling | "Query Store not enabled (unknown)" rendered red, auto-cleared at ~12s; no stale cancellation text anywhere all session |
| B3 compare state | Both buttons track window-wide plan count (headless-pinned; enabled state observed correct throughout) |
| B5 filter popup | Titled "Filter: Query Hash" |
| U1 empty state | Get started panel with real gestures + 5 recents; recent-click opened the plan; Connect row ran the real connect flow; first keystroke dismisses |
| U2 connect | Single click connected (no Test required); Database placeholder explains Test is optional |
| U3 double-click | Group rows expand, leaf loads the plan; no cell-edit flash (read-only grid) |
| U5 embedded toolbar | Plan sub-tab shows only plan-scoped controls; dead Reconnect/Database row gone |
| U6 tab sizing | Query Editor renders at sibling size |
| V3 formatter | Tooltips "3,475.5" / "4.7767"; properties "3,526.21" / "0"; QS grid IDs blank on grouped rows |
| V12 properties panel | Opens 380px; filter box narrows 60+ rows to a match and Esc restores; per-thread rollup collapsed by default with grouped metrics; probe residual renders full-width mono (no bracket tower); copy menu (no stock Paste); splitter drags and the width survives node switches |

**One more fix landed during the pass (`1e3be89`):** the column filter popup opened with a dead keyboard — Initialize called Focus() before the popup had a visual root. The dialogs workstream had flagged it as unverifiable-blind; verified live both ways (keystrokes went nowhere before, land in the value box now).

**Noted for round 2, not blocking:** the headless test suite writes into the real recent-plans store (worktree test paths showed up in the empty state's recents); the Query Store toolbar button keeps its active highlight when the view fails to open; toolbar still reflows between one and two rows across states (U4, phase 2 scope).

Final state: 26 commits, zero-warning solution build, 577 tests / 0 failed / 1 pre-existing skip.

https://claude.ai/code/session_017DEJ8FiSKJg44Ew2GG2nQ7

erikdarlingdata:
## Round 2 sweep

Second review pass into territory round 1 never entered. Two more fixes landed on this branch, both verified live before and after:

- **`0e328db` — reopening an open plan no longer stacks duplicate tabs.** The live session had accumulated three identical `another-parallel-spill.sqlplan` tabs; reopening now focuses the existing tab (matched on source path) and session restore heals duplicates.
- **`6f0a721` — the error dialog is dismissable.** It had no button and ignored Esc, and Alt+Tab could park it over the app indefinitely (observed with the pasted-plan XML error). Now: OK button, Enter/Esc both close, focus on the button.

**Documented for phase 2 (not fixed here):** tab-strip row wrapping is laptop-hostile — at 1280 logical width, 8 tabs + the toolbar made up to six rows of chrome; needs scroll + overflow menu alongside the U4 stable-toolbar work. The Query Store button's stuck accent fill (no toggle code exists — emergent template state) folds into the V6/U7 button-state redesign.

**Verified clean:** malformed-XML error content (line/position, no crash), 3MB plan loads ~6s with rendering intact, History pop-out/re-dock round trip, custom time range, Legend, Tab-key focus ring, small-window layout compresses without breaking, and the round-1 fixes re-confirmed in passing.

Gate: zero-warning build, 577 tests / 0 failed / 1 pre-existing skip. 28 commits total.

https://claude.ai/code/session_017DEJ8FiSKJg44Ew2GG2nQ7

PR #529, opened 2026-09-15
Move the minimap into a corner and settings out of About

erikdarlingdata:
Closes the last two open findings from the UI review, V7 and V11 — the two the campaign's phases never actually reached, despite the roadmap implying otherwise.

## V7 — the minimap

It opened as a **400×400 panel pinned to the canvas's top-left**. A plan renders left to right from its root, so that is exactly where the SELECT and the first operators are: the default covered the part you most want to see, and read as something to dismiss rather than navigate with.

Now **220×220, pinned bottom-right**, inset 18px so it clears the overlay scrollbar rails, with the panel opening *above* the toggle that summons it so neither covers the other. Resize still reaches 500 for anyone who wants the old size.

The toggle was the literal word `minimap` at 9px — a label, not a control. It is an icon button now. `AppIcons.Minimap` was drawn for this button during the phase 2 icon pass and left unwired, so the icon already existed; nothing referenced it.

Because the panel is pinned bottom-right its free corner is the top-left one, so the grip moved there and the drag math inverted. The drag is also re-based from `MinimapPanel`'s coordinates onto the control's: a panel pinned bottom-right moves its own origin as it grows, so in panel coordinates the grip would sit still while the pointer moved and the drag would fight itself.

Verified live: dragging toward the top-left grows it, the pinned corner stays put, and the plan's root is unobstructed.

## V11 — settings out of the About box

MCP server port and proxy configuration were in About. They are settings, not facts about the build, and nobody goes looking for a port number behind "About". They are now **Settings > Integrations**; About is back to version, copyright, links and the update check.

The section follows the dialog's Save/Cancel model rather than About's save-on-every-change, so **nothing reaches `settings.json` or the credential store until Save**, and both Reset Section and Reset All cover it. Unlike the other sections it holds its own pending state rather than being read back off its controls at save time, because those values must survive navigating away and back — which an orphaned control's value does not.

Verified live: the section renders, Manual reveals the proxy fields, changing the mode marks the dialog dirty, and discarding writes nothing (`proxy_mode` stays absent from settings.json).

## Two bugs found while doing it

**Saving Settings without visiting Query History silently wiped both of its values.** Sections build their controls only when first opened, and Save read them back unguarded — a null combo read as `""`, a null spinner as a hardcoded `10`. So the ordinary act of changing a Query Store option and pressing Save reset a configured max-plans of 50 to 10 and blanked the default metric. Format Options was already guarded against exactly this; Query History was not.

Confirmed by mutation rather than by reading: with the guard reverted the new test fails with `Expected: "TotalCpuMs", Actual: ""`.

**`SettingsFile` had no test-host redirect**, unlike `AppSettingsService` — which has one precisely because tests once rewrote the developer's real config. This change puts a save path to that file behind a UI a test can drive, so it is redirected now, before something writes a real `settings.json` and credential store from a test run.

## Checks

Four new tests; suite at **699 passing**, 1 skipped by design, 0 failures, 0 build warnings.

## Known, not in this PR

Editing a Settings section, navigating away and coming back loses the edit — the rebuilt controls re-read from the unchanged settings object while the dialog still believes it is dirty. The fix is to make the three existing sections write back on change the way the new Integrations section does. That is a refactor of all three and would make this PR hard to review, so it follows as its own.

https://claude.ai/code/session_017DEJ8FiSKJg44Ew2GG2nQ7


erikdarlingdata:
## Second pass: Avalonia review findings

An `avalonia-gotcha-reviewer` pass on the first commit found ten issues. Nine acted on in e1f4242; the two most serious were **regressions this branch introduced**.

**The resize grip covered the plan's root operator.** Pinning the panel bottom-right put its free corner — and so the grip — at the top-left, which is exactly where `PlanLayoutEngine` draws the root. The grip sets a transparent background and handles the press, so on any plan scaling below ~0.074 (about 23 leaves — routine) it swallowed the root's whole rect and click-to-center on it stopped working. The grip now lives in the panel header: same corner, no plan content underneath. Verified on screen — the map canvas is clear and the root is reachable.

**A lost pointer capture left the drag armed forever, and my sign flip made it far worse.** Nothing handled `PointerCaptureLost` and the move handlers only checked the dragging flag, so an alt-tab mid-drag left it set and a later plain *hover* resized against a stale origin. Before the inversion that shrank the panel to its floor; after, it grew it to 500 and swallowed the canvas. Both the grip and the canvas now clear on capture loss, plus a button-state bail in the move handler.

**Saving the proxy could delete a password nobody touched.** `ProxySettings.Load` swallows a credential-store failure and reports no password — indistinguishable from there being none — and the inherited rule ("touch unless the box is empty AND one is stored") then deleted it. Moving these settings widened the trigger from *a proxy field losing focus* to *any save*, so ticking the MCP checkbox was enough. The store is now touched only on positive instruction: a typed password, or a Reset asking for the stored one to go. Reset also says so now, instead of leaving the credential behind while the watermark claimed it was saved.

Also fixed: MCP and proxy writes split so one can't drag the other to the credential store; pending-state handlers now go through a guard that drops build-time events rather than silently overwriting the value just loaded; four handlers read their own control instead of a field reassigned on every rebuild; and Copy MCP command says so when the port isn't saved yet.

**Reset All wrote stale values back for every section not on screen** — the same root cause as the Query History bug already fixed here, and the review was right that it belonged in the same change. Pinned by a test that fails `Expected: 30, Actual: 90` without the fix.

Suite: **700 passing**, 0 failures, 0 warnings. The dirty-flag test now asserts the dirty flag and closes without leaving an unanswered modal in the shared headless app, and the file's note about credential-store safety now says what is actually true.

**Not taken:** the reviewer's note that `RenderMinimap()` runs on every pointer move during a resize. Real, but pre-existing and unchanged here.

Issue #548, opened 2026-09-21
ScrollBarVisibilityTests' DataGrid has no columns, so its rows never realize


erikdarlingdata:
Found during the Avalonia 12 upgrade: ADataGridsOwnScrollBarsFollowTheSameContract builds a 50-row DataGrid from anonymous items but never adds columns, and Avalonia's DataGrid does not auto-generate them — so zero rows realize and both PART_ scrollbars stay untemplated. The test's AllowAutoHide assertion is still meaningful (the value is template-bound, not assigned during OnApplyTemplate), and the doc comment now states the real mechanism, but the test could assert more if the grid actually overflowed: realized bars, a thumb, geometry under the app theme. Worth deciding whether to give it columns (changing what it pins) or leave it as the flag-only contract it documents itself to be.

erikdarlingdata:
Merged to dev in #554 — the flag-only contract test stays as documented, its construction comment now tells the truth, and a columns-bearing sibling pins realized bars keeping AllowAutoHide through the template binding. Ships in the next release.

Issue #547, opened 2026-09-21
The escape-brackets checkbox label loses its underscore to mnemonic parsing


erikdarlingdata:
QueryStoreGridControl.axaml:180: CheckBox Content="Treat [ ] _ as literals". AccessText takes the character after the first underscore as an access key and strips the underscore from display, so the label renders as "Treat [ ]  as literals" — losing the underscore that is half of what the checkbox is about (a LIKE wildcard). Pre-existing on Avalonia 11 and unchanged by 12 (found during the upgrade's access-key review). Fix: escape it as __ or reword the label.

erikdarlingdata:
Merged to dev in #553 — the underscore is escaped, the label renders whole. Ships in the next release.

Issue #546, opened 2026-09-21
Dispose the TextMate.Installation created for schema viewers


erikdarlingdata:
Found during the Avalonia 12 upgrade planning. The schema/definition viewers create a TextMate installation and never dispose it: PlanViewerControl.Schema.cs:114 and QuerySessionControl.Schema.cs:99. The query editor's own installation is disposed on detach (QuerySessionControl.axaml.cs), so the pattern exists in the codebase; the schema viewers just never adopted it. Pre-existing leak, not caused or worsened by Avalonia 12 (12.0.0's TextMate actually makes dispose/reinstall stricter via transformer ownership), which is why it was not fixed in the migration PR. Small fix: mirror the editor's dispose-on-detach.

erikdarlingdata:
Merged to dev in #550 — both schema viewers now run the query editor's install-on-attach/dispose-on-detach lifecycle, so closed tabs stop leaking tokenizer threads. Ships in the next release.

Issue #545, opened 2026-09-21
Remove the unused AvaloniaEdit.TextMate.Grammars package reference


erikdarlingdata:
Found during the Avalonia 12 upgrade planning. PlanViewer.App.csproj references AvaloniaEdit.TextMate.Grammars 0.10.12.1, but no app code uses it: the package ships its own RegistryOptions/ThemeName under the AvaloniaEdit.TextMate.Grammars namespace, while every call site binds TextMateSharp.Grammars instead (QuerySessionControl.Editor.cs uses it via a using directive; PlanViewerControl.Schema.cs fully qualifies it). The two never collide because the AvaloniaEdit namespace is never imported. Removing the reference also drops Newtonsoft.Json 13.0.1 and TextMateSharp 1.0.23 from the resolved graph, which is why it was kept out of the migration PR (a runtime-graph change in a release meant to be nothing but the migration). Safe cleanup for a normal release.

erikdarlingdata:
Merged to dev in #552 — the package reference is gone, and Newtonsoft.Json left the resolved graph with it. Ships in the next release.

PR #549, opened 2026-09-21
Upgrade to Avalonia 12

erikdarlingdata:
## Summary

Avalonia 11.3.22 → 12.1.2, with the coupled third-party releases: ScottPlot.Avalonia 5.1.59, Avalonia.AvaloniaEdit + AvaloniaEdit.TextMate 12.0.0, Avalonia.Controls.DataGrid 12.1.2, Avalonia.Headless 12.1.2. Migration only — no improvements rode along; the scope test applied throughout was "does Avalonia 12 make this wrong?"

**Do not merge until the manual pass below is done.** The suite proves a lot here, but three things are checkable only by hand.

## App changes (the whole compile-and-runtime surface)

- **Two one-line API migrations, both breaking item #13:** `DataGridBehaviors.cs` moves off the removed `TopLevel.PlatformSettings` onto `grid.GetPlatformSettings()` (with a new in-suite `Assert.NotNull` guard, since Windows Ctrl+C cannot falsify it — the fallback is also Ctrl). And `QuerySessionControl.GetParentWindow()` moves off `VisualRoot as Window` — 12 hosts the Window under a new `TopLevelHost`, so the old form compiled, ran, and returned the wrong thing silently, which would have broken every session-owned dialog. That one no compile probe could catch; the headless suite did.
- 13 `Watermark` → `PlaceholderText` renames (12 XAML + 1 C#), plus the same file's parameter and prose stragglers. Visual parity is settled by resource-set identity (Fluent ships the identical placeholder resources in 11 and 12; only unreferenced `PART_` names changed).
- A vestigial CS0618 pragma deleted; the ScottPlot version-gate comment deleted; the SkiaSharp pin comment rewritten to say what the pin actually does under 12 (nothing — it is kept because the #139 NoDependencies collision is unresolved and removing it is out of scope); the DataGrid `AllowAutoHide` comment updated for 12's template-binding mechanism (upstream PR #241).
- Zero code changes for ScottPlot, AvaloniaEdit, and DataGrid usage — every API the app touches survives, verified against the shipped assemblies.

## Test suite

713 tests at baseline, 713 at tip — zero deleted, zero added, no new skips, no assertion softened (both exact-collection toolbar assertions stayed exact collections). The measurement churn comes from 12 replacing the headless stub text shaper (flat 10 DIP/char) with real HarfBuzz over an embedded font: widths now scale by FontSize/10, line heights by a flat ×1.36. That cause is written once, on `HeadlessUi`, and every re-measured threshold points at it. The toolbar's protected-command invariant (Connect/Execute/Est never collapse) is now stated explicitly in the assertions that guard it.

**Harness:** kept hand-rolled (12's xunit-v3 support removes the original reason but not the conclusion — the #474 drain and the survival canary don't exist in `[AvaloniaFact]`). The #474 note now says precisely what 12 fixed (the scope leak) and didn't (the throw — the drain remains the only prevention). A teardown-failure precedence guard shipped after an empirical probe confirmed 12 routes teardown throws into the dispatch task, where they were replacing assertion messages.

**#544 — Avalonia 12's own headless setup has a thread-affinity race** (~1 run in 4, one arbitrary victim, zero repo frames on the stack, not reproduced on 11, not caused by test concurrency — parallelization-off still hit it). Mitigated with a setup-scoped single re-dispatch that structurally cannot retry a test: it fires only when the dispatched delegate provably never started AND the stack matches Avalonia's setup frames, writes to stderr on every occurrence, and documents its removal condition. Proven engaging on the real race: the marker fired 3 times across 20 test-host runs, none failed. Gate: 10 consecutive complete green `dotnet test` runs, plus 2 more by the reviewing orchestrator.

## Verification

- Release build 0 errors / 0 warnings; 12 consecutive complete green suite runs across two operators.
- linux-x64 self-contained publish succeeded; published `libSkiaSharp.so` SHA256-matches `SkiaSharp.NativeAssets.Linux 3.119.4` and differs from ScottPlot's NoDependencies copy — issue #139 does not reproduce on this graph.
- Planned and reviewed by a multi-agent pipeline: 3 planners (one ran a full compile probe against real 12.1.2), 4 adversarial reviewers over 3 rounds (18 verified Major findings folded), 2 bucket-(b) escalations ruled mid-build. Plan and rulings are in the local crew doc.

## Before merge (manual, ~20 minutes, installed or clean Release build only)

The twelve-point pass in the plan doc, headlined by the three things the suite structurally cannot see: loading-spinner animation (12 newly completes style animations on detach — undocumented), the #464 scrollbar rail geometry, and AvaloniaEdit highlighting across repeated tab switches (the 12.0.0-editor-on-12.1.2-core skew). Plus Alt mnemonics (12 matches layout symbols now) and the clipboard guards.

## Follow-ups filed

#544 (the race, with removal condition), #545 (drop unused Grammars package), #546 (TextMate installation leak in schema viewers), #547 (pre-existing mnemonic eating the escape-brackets label's underscore), #548 (the columns-less DataGrid test). Before the first 12-based release: confirm SignPath's 'App' artifact configuration signs by glob — `Avalonia.HarfBuzz.dll` is new in the publish set. After merge: dispatch nightly.yml once and eyeball the linux artifact.

macOS double-click activation remains the standing open question; it rides with the next Mac session, now alongside the Option-mnemonic check.

https://claude.ai/code/session_01Sr43FouJ6RizUyoWQeR5FX


erikdarlingdata:
Post-review addition, pushed as 12f64dc: the driven GUI pass executed the plan's spinner check and hit the pre-classified bucket-(b) item exactly where the regression review predicted — Avalonia 12 completes style animations on detach (undocumented, outside the PlaybackBehavior guard), so a loading bar frozen mid-track after any tab switch during a capture. Fixed with ProgressBarBehaviors.RestartOnReattach applied to all seven at-risk bars; proven both directions in driven runs (five frames across a full animation cycle: pixel-identical before, sweeping after). Suite re-run green at 713. The drive also verified live on this branch: connection dialog through the fixed GetParentWindow path, the #542 database handover, placeholders, Alt mnemonics with underlines, editor typing/colorization/completion, clipboard copy via Get-Clipboard, scratch restore, and the error path. Items still needing human hands are unchanged: scrollbar rail geometry, chart box-select vs pan, middle-click pan, grid stripe under row recycling, sorts/resize. One pre-existing (11-reproduced) find filed separately: #550-adjacent focus-after-Ctrl+N, issue #551.

erikdarlingdata:
SignPath check from the review (regression MIN1) is resolved — no action needed. The 'App' artifact configuration signs exactly one named PE, PlanViewer.App.exe, inside the zip; the second config signs the MSI container. Neither enumerates the payload, so the new Avalonia.HarfBuzz.dll passes through unsigned exactly like every Avalonia/SkiaSharp DLL already does in every shipped release — the strongest proof being history itself: releases have always contained dozens of unlisted DLLs, so unlisted files demonstrably do not fail the request. Avalonia 12 renames neither the exe nor the MSI. Nothing on the signing side gates this merge.

Issue #556, opened 2026-09-23
[BUG] Non-SARGable Predicate (rule 12): false positive when a parameter-side function follows another comparison in a compound predicate


nmummau:
### Component

CLI Tool

### Performance Studio Version

1.26.0

### Operating System

Windows 11

### Describe the Bug

Rule 12 (Non-SARGable Predicate) reports Function call (CONVERT) on column when the CONVERT wraps a parameter, not a column. It happens when the predicate has more than one comparison joined by AND or OR.

IsFunctionOnColumnSide in PlanAnalyzer.Detection.cs finds the first comparison operator in the whole predicate and splits there. It then checks whether the side with the function contains a column reference:

var compMatch = Regex.Match(predicate, @"(?<![<>])([<>=!]{1,2})(?![<>=])");
...
string side = funcSide == "left" ? predicate[..compPos] : predicate[(compPos + compMatch.Length)..];
return ColumnReferenceRegex.IsMatch(side);

In the predicate below, the first = is after [t].[A]. So the "function side" of CONVERT(tinyint,[@2],0) is everything to the right of that first =, which includes the second condition's column [t].[B]. The column check matches, and the function is treated as column-side. #437 (for #436) fixed the equivalent case for CONVERT_IMPLICIT, but IsFunctionOnColumnSide still uses one split point for the whole predicate.

Version: v1.26.0 (e6593df), planview analyze with the default rules, on SQL Server 2022 (16.0.4195.2).

### Steps to Reproduce

1. Create a table with no index on the filtered columns:
CREATE DATABASE Repro;
GO
USE Repro;
CREATE TABLE dbo.T (Id INT IDENTITY PRIMARY KEY, A INT NOT NULL, B TINYINT NOT NULL);
INSERT dbo.T (A, B) VALUES (52, 2), (53, 1);
2. Run this query and save its actual execution plan (attached as non-sargable-compound-predicate.sqlplan):
SELECT COUNT(*) FROM dbo.T AS t WHERE t.A = 52 AND t.B = CONVERT(tinyint, 2);
   SQL Server auto-parameterizes it as (@1 tinyint,@2 int)SELECT COUNT(*) FROM [dbo].[T] [t] WHERE [t].[A]=@1 AND [t].[B]=CONVERT([tinyint],@2). The plan is a Clustered Index Scan with this predicate:
[...]

Issue #436, opened 2026-08-20
[FEATURE] Distinguish between severity of Warnings


pgfiore:
### Which component(s) does this affect?

- [x] Desktop App
- [ ] CLI Tool
- [ ] SSMS Extension
- [x] Plan Analysis Rules
- [ ] Documentation

### Problem Statement

Since English isn’t my native language, I'll start with a use case to make my life easier.

A query in SSMS (say, the latest 22) doesn't show any specific message, but in Performance Studio (PS) it shows the warning below. 

<img width="356" height="271" alt="Image" src="https://github.com/user-attachments/assets/ac37bab1-8be4-4ec7-bd2b-d93ac3f699e7" />

I can only find one a.b.normal message; a table scan shows the predicate: dbo.IB03_GARANZ.UNIV_SEZIONI as IB03.UNIV_SEZIONI=CONVERT_IMPLICIT(numeric(18,0),[@0],0) (and @0 is the forced parametrization at work). And in SSMS, there’s no yellow exclamation mark to highlight warnings. 

I have two doubts about PS behavior:

1) This is not a SARGable issue. The IB03.UNIV_SEZIONI is numeric(), which has a highest precedence on int. And the column is searchable (*).

2) I like PS showing “hidden” issues, but I’d like to have a way to differentiate between SQL Server warnings and PS help.

The plan is enclosed. Thank you for being so helpful.

(*) I tried the other way around and, in fact, "cast(IB03.UNIV_SEZIONI as varchar) = 5004001" correctly fires a warning in SSMS too.

### Proposed Solution

Fix warnings severity
And differentiate between SQL Server warnings and PS "help"

### Use Case

drop table if exists #IB03_GARANZ
create table #IB03_GARANZ (UNIV_PK numeric(18,0) primary key, UNIV_SEZIONI numeric(18,0))
select * from #IB03_GARANZ IB03 where UNIV_SEZIONI=1

erikdarlingdata:
You're right on point 1, and it's a bug in the tool, not a subtlety you're missing. Your English is fine — the SQL reasoning is better than fine.

**Why the warning is wrong.** Data type precedence decides which side SQL Server converts, and it converts the *lower*-precedence side. `numeric(18,0)` outranks `int`, so comparing your column to an integer converts the **parameter** up to the column's type, which is exactly what the plan says:

```
[...]

PR #557, opened 2026-09-23
Read only the function's own comparison in the non-SARGable check

erikdarlingdata:
## Summary

Fixes #556. Rule 12 (Non-SARGable Predicate) reported a function on a column when the function wrapped a parameter. This happened when the predicate had more than one comparison joined by AND or OR.

The side check split the whole predicate at its first comparison operator. In a compound predicate, that put the next comparison and its column on the same side as the function. The check now reads only the comparison that contains the function. It splits the predicate at every AND and OR, also inside parentheses. It does not split inside a string literal or a bracketed name.

The same mistake also flagged a common date range: `[d]>=dateadd(day,(-7),getdate()) AND [d]<getdate()`. This shape is probably the most frequent trigger in real plans.

## Also fixed

Two related false positives in the same rule:

- The ISNULL/COALESCE check flagged every ISNULL, on either side. It now uses the same side check. `col = ISNULL(@p, 0)` is no longer flagged. `col = ISNULL(@p, col)` is still flagged, because the column is inside the function.
- The side check did not treat LIKE as a comparison. So `[col] like upper([@p])` fell back to the worst case and was flagged. LIKE is now a comparison operator for this check.

## Tests

- The reporter's plan is a new fixture. Its test asserts that no Non-SARGable warning appears. It also asserts that Scan With Predicate still appears, because that scan is real.
- New tests run on raw predicate strings. They cover the date range, a parenthesized OR group, AND inside a string literal, LIKE on both sides, and three ISNULL shapes. `DetectNonSargablePattern` is now internal so that these shapes can be tested without a plan file. #436 did the same for `ConvertImplicitWrapsColumn`.
- Each new test fails when its fix is removed. With the old split of the whole predicate, five tests fail. With an AND split that ignores string literals, the literal test fails. Without LIKE as an operator, the LIKE test fails.
- The two golden-master files change only because of the new fixture. The warning digest gets one new section. The comparison digest gets two new pairings, and one existing pairing now points at the new fixture. The output for every other plan is the same.
- `planview analyze` on the reporter's plan now reports one warning, Scan With Predicate.
- Full suite: 723 passed, 1 skipped (the known macOS test), 0 failed.

https://claude.ai/code/session_01E9G1hWs6Nk4h4U2gn7TSy5


nmummau:
@erikdarlingdata Thanks for the quick turn around. When do you plan on cutting the next release?

